PR #25, opened 2019-02-17
Frontend refactoring

christiaua:
Added hocs (Higher order components) for HeaderBar, refactored directories for screens and added a Switch in the Router


client/src/screens/Login/Login.js
```diff
@@ -0,0 +1,25 @@
+import React, { Component, Fragment } from "react";
+import "./Login.css"
+
+class Login extends Component {
+	render() {
+		return (
```
patlai: nit: keep tabbing consistent between files

PR #26, opened 2019-02-17
Adding student user removal

si-x-li:


Ty-Won:
On hold, due to styling issues with promises,

Ty-Won:
@si-x-li Nice job! A few things to note, 

1)Is there a formatting tool you used? (i.e like prettify that Kevin suggested so that it makes the code look more uniform?)
2) The code you wrote seems very thourough about handling errors (Which is fantastic :+1: ) Have you investigated the use of .catch for promises? I'm just trying to see if that may save you a few lines and make error handling a bit more tidy

si-x-li:
1) No formatting tool was used but I downloaded VSCode. I'll format my code next time before PR
2) .catch would tidy up the code linearly, but the current structure with callbacks allows me to log and return a custom error message rather than a generic one

Example: 
try {
  await someAsyncFunction1();
  await someAsyncFunction2();
} catch (err) {
  console.error(err);
  return reject/resolve(false);
}
resolve(true);

Here, I can only reject or resolve with a generic message so I don't know what's wrong. It could be that someAsyncFunction1() is inserting into users table and someAsyncFunction2() and that someAsyncFunction2() fails due to duplicate keys. I'm only getting false

Example with callbacks:
someFunction1(function(error, value) {
  if (error) {
    let msg = 'Username is already taken';
    console.err(msg);
    reject(msg);
  }
  someFunction2(function(error, value) {
     if (error) {
       let msg = 'Student ID is already taken';
       console.err(msg);
       reject(msg);
     }
     resolve(true);
  }
}

Currently, the error handling is implemented as a generic but we should discuss about it tonight

Ty-Won:
Woah, thanks for the thorough explanation, you do great work as usual steven, keep it up! Hmm I see you're using async, I believe that promises and async can be used interchangeably, so like for example 
```Node
await someAsyncFunction1().then(function(result){
             ...
}.catch(function(error){
             console.error(error);
});

await someAsyncFunction2().then(function(result){
             ...
}.catch(function(error){
             console.error(error);
}
```

this i just a nitpick btw, if your code works and others can read it, its good enough for me! I was just wondering about your reasoning behind the chained callbacks

MartyO256:
To improve on the readability of the tests, I would add a function that performs the assertion after receiving the response from the database.
```js
  const testResponse = async (
    { username, password, email, id },
    expectedResponse,
  ) => {
    const actualResponse = await users.insertStudentUser(
      username,
      password,
      email,
      id,
    );
    assert.equal(actualResponse, expectedResponse);
  };
```
This way, test cases can be executed like this.
```js
  it.only("responds with undefined username", () =>
    testResponse(
      {
        username: null,
        password: "password",
        email: "email@email.com",
        id: "123456789",
      },
      "undefined username",
    ));
```
This would allow for test cases and expected responses to be stored separately in a JSON file. Then [dynamically generated tests](https://mochajs.org/#dynamically-generating-tests) could be made for those, which would greatly shorten the test script.

Ty-Won:
@MartyO256 Nice suggestion! Seems like you know your stuff :+1: Thanks for the suggestion, always looking for ways to improve this project :') I encourage your contributions to other open source projects, it's great what you are doing, keep it up! Just a fair warning, this isn't really an "open sourced project", it's being done for a class design project atm so we won't be able to accept/credit you for any contributions (At least as far as I know lol)

si-x-li:
Moved to #32

PR #28, opened 2019-02-18
Added API for retrieving user completed courses

KevenLYF:
1. Added API for retrieving user completed courses
2. Formatted some files with Prettier
3. Restructured test folder

### DIFF
--- a/backend/logic/users/users.js
+++ b/backend/logic/users/users.js
@@ -1,8 +1,10 @@
+const util = require("util");
+
 const mysql = require("../../sql/connection");
 const format = require("../../validation/format");
 const hasher = require("../../validation/hash");
 
-exports.insertStudentUser = (username, password, email, id) => {
+var insertStudentUser = (username, password, email, id) => {
   return new Promise(function(res, rej) {
     // Connect to database
     let promise = new Promise(function(resolve, reject) {
@@ -107,9 +109,37 @@ exports.insertStudentUser = (username, password, email, id) => {
   });
 };
 
+var getCompletedCourses = async studentID => {
+  let courses = [];
+
+  const sql_query = `SELECT course_code, semester
+  		FROM course_offerings
+  		WHERE (id, semester)
+  		IN (SELECT offering_id, semester FROM student_course_offerings WHERE student_id = ?);`;
+
+  mysql.query = util.promisify(mysql.query);
+
+  let results;
+  try {
+    results = await mysql.query(sql_query, [studentID]);
+  } catch (err) {
+    console.log(err);
+  }
+
+  if (results) {
+    courses = JSON.stringify(results);
+  }
+  return courses;
+};
+
 function logError(connection, error) {
   console.error(error);
   connection.rollback();
   connection.release();
   return "false";
 }
+
+module.exports = {
+  insertStudentUser,
+  getCompletedCourses
+};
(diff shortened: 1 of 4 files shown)


backend/routes/users.js
```diff
@@ -1,18 +1,39 @@
...
+ * 		curl -X GET -H "Content-Type: application/json" 'http://localhost:3000/users/getCompletedCourses?studentID=12345'
+ *
+ * @returns An json of completed course code, e.g [{"course_code":"ECSE422","semester":"W2019"},{"course_code":"ECSE428","semester":"F2019"}]
+ *
+ * @author: Yufei Liu
+ *
+ */
+router.get("/getCompletedCourses", async (req, res) => {
+  const student_id = req.query.studentID;
+  const courses = await users.getCompletedCourses(student_id);
```
Ty-Won: Loving the formatting and the detailed specifications on the top! Nice job! I looked at the getCompleteCourses, what happens if there is an error? Will this route still return status 200?

KevenLYF: Yeah it will still send 200, but with an empty output "[]", or we can make it send status 5xx. Just trying to make it simple for now. Good point, I'll fix it later!

si-x-li:
Let #31 merge first, then merge this PR

KevenLYF:
ok

backend/routes/users.js
```diff
@@ -1,18 +1,39 @@
...
+ * @apiExample {curl} Example usage:
+ * 		curl -X GET -H "Content-Type: application/json" 'http://localhost:3000/users/getCompletedCourses?studentID=12345'
+ *
+ * @returns An json of completed course code, e.g [{"course_code":"ECSE422","semester":"W2019"},{"course_code":"ECSE428","semester":"F2019"}]
+ *
+ * @author: Yufei Liu
+ *
+ */
+router.get("/getCompletedCourses", async (req, res) => {
+  const student_id = req.query.studentID;
```
AlexLam1997: Consider using req.body to allow for JSON to be passed through the body of the request

KevenLYF: req.body is used to `POST` or `PUT` request. Since those requests needs to pass data as a body.

backend/routes/users.js
AlexLam1997: Can we make it send back an actual error when it fails here?

backend/logic/users/users.js
```diff
@@ -107,9 +109,37 @@ exports.insertStudentUser = (username, password, email, id) => {
...
+  		WHERE (id, semester)
+  		IN (SELECT offering_id, semester FROM student_course_offerings WHERE student_id = ?);`;
+
+  mysql.query = util.promisify(mysql.query);
+
+  let results;
+  try {
+    results = await mysql.query(sql_query, [studentID]);
+  } catch (err) {
+    console.log(err);
```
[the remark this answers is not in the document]
KevenLYF: yes good point! Actually I was planning to add more error handling in another patch. (I thought we want to merge this one ASAP and then merge other PRs)

KevenLYF:
Yes we definitely should have more Error handling here (like 500 or 400). I was planning to make it another patch. So we merge this one ASAP so that other PRs won't be blocked. Then I'll add a new patch to fix the error handling

backend/logic/users/users.js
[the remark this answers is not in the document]
AlexLam1997: makes sense, in that case, approved!

PR #29, opened 2019-02-18
Alex/backend/tests/modify tests

AlexLam1997:
Didn't see stevens remove user method at the time this was written. 
This has:
-Remove method as promise
-Implements database queries as promises instead of callbacks
-Changes in the user tests to delete the inserted entries after the inserts are made


backend/logic/courses/courses.js
```diff
@@ -0,0 +1,9 @@
+const mysql = require('../../sql/connection');
+
+exports.queryCoursesByTag = (tag) => {
+    return new Promise(function(resolve,reject){
+        let promise = new Promise(function(resolve,reject){
```
Ty-Won: I like that you are using promises, thats good! I think its important to know why you used promises here. Like for instance, you have promises with the resolve and rejection, so feel free to use them to indicate how your code will proceed after the completion of an asynchronous operation

backend/logic/users/users.js
```diff
@@ -1,13 +1,17 @@
...
 		// Connect to database
-		let promise = new Promise(function(resolve, reject) {
-			mysql.getConnection(function(err, conn) {
-				if (err) { reject(err); }
+		let promise = new Promise(function (resolve, reject) {
+			mysql.pool.getConnection(function (err, conn) {
+				if (err) {
+					reject(err);
+				}
 				resolve(conn);
```
Ty-Won: Great job handling the error, I believe this is the way it is done! Just a quick concern that I'm not 100% sure about, does code execute after reject(err) is called? If so, is there potential for it to execute resolve(conn) in this case?

backend/logic/users/users.js
```diff
@@ -61,22 +65,31 @@ exports.insertStudentUser = (username, password, email, id) => {
...
-						// Commit transaction
-						connection.commit(function(error) {
-							if (error) { res(logError(connection, error)); return; }
-							connection.release();
-							res('true');
-						});
+					[username, email, hash],
+					function (error, results, fields) {
+						if (error) {
+							res(logError(connection, error));
```
Ty-Won: Again good to see you're looking to handle errors, it makes your code more solid. Just noticed that you used resolve instead of reject, was this intentional?

backend/logic/users/users.js
```diff
@@ -86,6 +99,26 @@ exports.insertStudentUser = (username, password, email, id) => {
...
+	try {
+		let query = 'SET SQL_SAFE_UPDATES = 0; DELETE FROM students WHERE students.username LIKE ?;DELETE FROM users WHERE users.username LIKE ?; SET SQL_SAFE_UPDATES = 1;'
+		await connection.beginTransaction();
+		await connection.query(query, [username, username]);
+		await connection.commit();
+	} catch (error) {
+		console.error(error)
+		connection.rollback();
+		connection.release();
+	}
```
Ty-Won: To be discussed with steven over which method you guys will be using

Ty-Won:
@AlexLam1997 Nice job! I've left some small comments over some detals I've notice, @patlai @KevenLYF  node experts please add in anything I might have missed

KevenLYF:
Haha thanks altho I'm not a Node expert xd. I'll when I have time!

backend/logic/courses/courses.js
AlexLam1997: Oh my bad, this is incomplete code that is completed in my other PR

AlexLam1997: Thanks for catching this, it will however be resolved after the other one is merged

backend/logic/users/users.js
AlexLam1997: Nope, reject makes the promise "return" an error if i am not mistaken

backend/logic/users/users.js
AlexLam1997: This is one of the things about error handling to discuss during the meeting.  Steven suggested resolving with a status message. Other options also being considered. Will be brought up during the meeting

backend/logic/users/users.js
KevenLYF: Tyrone is right, I think the code after 'reject' will get called. (In this case it's resolve). However, the call `resolve(conn)` won't do anything because the promise it's already been fulfilled. Once a promise is fulfilled, later `resolve` or `reject` call won't have any effects on it. (I might not be 100% correct, better to check the doc)    So generally it's ok to leave it like this, but put a return is definitely a better practice :)

backend/sql/connection.js
```diff
@@ -1,14 +1,40 @@
...
-	database: 'heroku_f11d5f6ef0dc063'
+	database: 'heroku_f11d5f6ef0dc063',
+	multipleStatements: true
+
+	//PROD database 
+	/*
+	connectionLimit: 100,
+	host: 'us-cdbr-iron-east-03.cleardb.net',
+	user: 'bc3b1bec1b7226',
+	password: 'b6054e24',
```
MartyO256: I don't think the production database credentials should have ended up on Github.

Ty-Won: @MartyO256 Thanks for that, I commented this in another pull request. I appreciate the contribution! How did you find yourself here lol

PR #30, opened 2019-02-18
added course retrieval code with unit tests and additional db insert scripts

AlexLam1997:
Depending on the outcome of the callback/promise/async discussion this may cause conflicts that will need to be resolved when merging the removeUser and modifyTest branches. 

However this branch affects very little compared to those two, and should be merged in first


backend/logic/courses/courses.js
```diff
@@ -0,0 +1,10 @@
+const mysql = require('../../sql/connection');
+
+async function queryCourseByTag(tag){
+    //if an invalid tag is passed, an empty object is returned
+    let connection = await mysql.getNewConnection();
+    let courses = await connection.query('SELECT course_code FROM course_tag WHERE course_tag.tag_name LIKE ?', tag);
+    return (JSON.parse(JSON.stringify(courses)));
```
si-x-li: The connection needs to be released before you return courses

backend/sql/connection.js
```diff
@@ -1,14 +1,40 @@
...
 
 let pool = mysql.createPool(db_config);
 
-module.exports = pool;
+async function getNewConnection() {
+	try {
+		pool.getConnection = util.promisify(pool.getConnection);
+		conn = await pool.getConnection();
+		conn.query = util.promisify(conn.query);
+		conn.beginTransaction = util.promisify(conn.beginTransaction);
```
si-x-li: Transaction is not needed for every type of query

AlexLam1997: this is just to set up the call to "beginTransaction" to return a promise

backend/logic/courses/courses.js
AlexLam1997: Modified, thanks!

backend/sql/connection.js
si-x-li: Got it looks good then

backend/logic/courses/courses.js
```diff
@@ -0,0 +1,10 @@
+const mysql = require('../../sql/connection');
+
+async function queryCourseByTag(tag){
```
si-x-li: Forgot to mention, verify that tag is not undefined

AlexLam1997: done!

backend/sql/connection.js
```diff
@@ -1,14 +1,40 @@
...
+		return(conn);
+	} catch (error) {
+		console.log(error);
+	}
+}
+
+module.exports.pool = pool;
+module.exports.getNewConnection = getNewConnection;
```
KevenLYF: There is a short-hand expression
`module.exports = {
  pool,
  getNewConnection
};`

You can check my commit in logic/user.js

backend/sql/connection.js
```diff
@@ -1,14 +1,41 @@
...
+
+	//PROD database 
+	/*
+	connectionLimit: 100,
+	host: 'us-cdbr-iron-east-03.cleardb.net',
+	user: 'bc3b1bec1b7226',
+	password: 'b6054e24',
+	connectionTimeout: 10000,
+	database: 'heroku_35d3c28bdd7d221'
+	*/
```
Ty-Won: this is just a nitpick, I know our app isnt like professional use ^^ but lets hide the credentials to our DB

Ty-Won:
Great Job, overall looks nice to me. We will discuss this in the meeting

backend/sql/connection.js
AlexLam1997: Hmm this would be tough to do and involve an additional encryption layer but I think I can come up with something, would need to add an additional task in the task list though as its definitly going to take me some time

KevenLYF: We can store the DATA base credentials in our local environment as Environment Variables, and then use sth like `process.env.ENV_VARIABLE` to get the credentials. But I think we can put this as low priority. As if we do this, everyone needs to store the credentials on their local environment.

backend/logic/courses/courses.js
```diff
@@ -1,11 +1,19 @@
...
-    await connection.release();
-    return (JSON.parse(JSON.stringify(courses)));
+    try {
+        let courses = await connection.query('SELECT course_code FROM course_tag WHERE course_tag.tag_name LIKE ?', tag);
+        connection.release();
+        return (JSON.parse(JSON.stringify(courses)));
+    } catch (error) {
+        connection.release();
+        console.error(error);
+        throw new Error(error.message);
```
Ty-Won: @si-x-li is curious about the throwing of the error, does something catch this?

backend/routes/users.js
```diff
@@ -4,10 +4,10 @@ const router = express.Router();
 
 /* GET users listing. */
```
KevenLYF: Could you add some comments here? Like an explanation for the API, you can reference my format in PR #28

AlexLam1997: done, thanks for the pointer

KevenLYF: Welcome :)

backend/logic/users/users.js
```diff
@@ -1,112 +1,48 @@
...
-
-        if (error) {
-          connection.release();
-          console.error(error);
-          res(error);
-          return;
-        }
+var insertStudentUser = async (username, password, email, id) => {
+  // Connect to database
+  let connection = await mysql.getNewConnection()
```
si-x-li: getNewConnection() can be moved to the try catch block

backend/logic/users/users.js
```diff
@@ -1,112 +1,48 @@
...
+    error = "invalid format email";
+  } else if (!format.verifyId(id)) {
+    error = "invalid format id";
+  }
 
-        // Hash the password
-        let hash = hasher.hashPass(password);
+  console.error(error)
+  if (!error == false) {
+    connection.release();
```
si-x-li: connection.release() can be moved to the try catch block as well

backend/logic/users/users.js
AlexLam1997: thats true thanks!

si-x-li:
Looks good from me

PR #31, opened 2019-02-20
Added env variables for DB credentials and fixed some formatting

Ty-Won:
Changed the DB credentials to use env variables, I will post the .env file on slack later so that everyone can use this.

### DIFF
--- a/backend/logic/users/users.js
+++ b/backend/logic/users/users.js
@@ -1,94 +1,115 @@
-const mysql = require('../../sql/connection');
-const format = require('../../validation/format');
-const hasher = require('../../validation/hash');
+const mysql = require("../../sql/connection");
+const format = require("../../validation/format");
+const hasher = require("../../validation/hash");
 
 exports.insertStudentUser = (username, password, email, id) => {
-	return new Promise(function(res, rej) {
-		// Connect to database
-		let promise = new Promise(function(resolve, reject) {
-			mysql.getConnection(function(err, conn) {
-				if (err) { reject(err); }
-				resolve(conn);
-			});
-		});
+  return new Promise(function(res, rej) {
+    // Connect to database
+    let promise = new Promise(function(resolve, reject) {
+      mysql.getConnection(function(err, conn) {
+        if (err) {
+          reject(err);
+        }
+        resolve(conn);
+      });
+    });
 
-		// Await for connection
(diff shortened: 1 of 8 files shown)


backend/routes/users.js
```diff
@@ -1,17 +1,30 @@
...
+router.post("/completedCourses/comparison", function(req, res) {
+  let completedCourses = req.body;
+  curriculum
+    .courseComparison(completedCourses)
+    .then(function(remainingCourses) {
+      res.send(remainingCourses);
+    })
+    .catch(function(err) {
+		console.error(err);
+		res.send(/*TODO ERROR CODE*/);
```
si-x-li: TODO code left here

backend/logic/users/users.js
```diff
@@ -1,94 +1,115 @@
-const mysql = require('../../sql/connection');
```
si-x-li: Possible merge conflicts with #28. Will need to be resolved

backend/routes/users.js
```diff
@@ -1,17 +1,30 @@
...
-	let email = req.query.email;
-	let id = req.query.student_id;
-	users.insertStudentUser(username, password, email, id)
-	.then(function(val) {
-		console.log(val);
-		res.send(val);
+router.post("/addStudentUser", function(req, res, next) {
+  let username = req.body.username;
+	let password = req.body.password;
+	let email = req.body.email;
```
si-x-li: Formatting is broken

backend/logic/users/users.js
Ty-Won: is it because of the single quotes? 0-0

si-x-li: users.js is formatted in #28 and this

KevenLYF: we can merge this first, then I'll fix my commit #28, then we merge #28

backend/routes/users.js
KevenLYF: we can send 500 it it's a server error (Backend error)

backend/routes/users.js
KevenLYF: once we fixed this formatting we can merge this

Ty-Won:
@si-x-li @KevenLYF I chose to force update since the changes were very small and minor, no need to create extra commit history with no purpose

PR #33, opened 2019-02-21
Gareth/backend/login

GarethDPeters:
Addition of login functionality for both users with passport. Has not been tested yet.


backend/logic/users/users.js
```diff
@@ -67,7 +74,77 @@ var getCompletedCourses = async studentID => {
...
+			await connection.beginTransaction();
+			let userInfo = await connection.query('SELECT * FROM users WHERE username = ?;',[username]);
+			await connection.commit();
+			connection.release();
+
+			if(!userInfo) {
+				return done(null, false, { message: 'Incorrect username.' });
+			}
+
+			if (!isValidPassword(userInfo[0].password, password)) {
```
si-x-li: Since all data in the database is hashed using the crypto package and you just fetched all user information from the database, there is no need to compare with the bcrypt package

1) Hash input password with the hasher function in hash.js
2) Compare

backend/logic/users/users.js
```diff
@@ -26,7 +28,12 @@ var insertStudentUser = async (username, password, email, id) => {
   }
 
   // Hash the password
-  let hash = hasher.hashPass(password);
+	var generateHash = function(password) {
+		return bCrypt.hashSync(password, bCrypt.genSaltSync(8), null);
+	};
+
+	let hash = generateHash(password);
```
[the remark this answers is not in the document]
AlexLam1997: This method is to insert a new student though, so i think its fine, we would however meed to store the salt in order to autheticate the password when they try to login

backend/logic/users/users.js
```diff
@@ -67,7 +74,77 @@ var getCompletedCourses = async studentID => {
...
+		usernameField : 'username',
+		passwordField : 'password',
+		passReqToCallback : true // allows us to pass back the entire request to the callback
+	},
+	
+	function(req, username, password, done) {
+		let error;
+
+		let isValidPassword = function(userpass, password){
+			return bCrypt.compareSync(password, userpass);
```
AlexLam1997: Just curious, How does bceypt compare without a salt to rehash agaisnt?  Like when you hashed tbe password, you needed to hand in a salt to create the hash, doesnt it need that same salt in order to decrypt the password in the database?

backend/logic/users/users.js
```diff
@@ -167,10 +169,76 @@ var getCompletedCourses = async studentID => {
...
+	app.passport.use('local-signin', new LocalStrategy(  
+	{
+		// by default, local strategy uses username and password, we will override with email
+		usernameField : 'username',
+		passwordField : 'password',
+		passReqToCallback : true // allows us to pass back the entire request to the callback
+	},
+	
+	function(req, username, password, done) {
+		let error;
```
patlai: you can refactor all the error checking logic into a separate function to clean this one up a bit

backend/logic/users/users.js
```diff
@@ -167,10 +169,76 @@ var getCompletedCourses = async studentID => {
   return courses;
 };
 
+var login = async () => {
```
patlai: nit: inconsistent tabbing (use either 2 or 4 spaces but not both in the same file)

KevenLYF: @GarethDPeters  You are using Prettier right? you can try turn on its format-onsave feature

backend/logic/users/users.js
```diff
@@ -167,10 +169,76 @@ var getCompletedCourses = async studentID => {
...
+				return done(null, false, { message: 'Incorrect password.' });
+			}
+
+			return done(null,userInfo[0]);
+		} 
+		catch (error) {
+			connection.release();
+			console.error(error);
+			return done(error);
+		}
```
Ty-Won: Killing it bud :') just a small suggestion, I notice that you have connection.release() twice in the try catch, you could alternatively do:

```node
try{
  ...
}
catch{
  ...
}
finally{
  connection.release()
}
```

backend/logic/users/users.js
```diff
@@ -167,10 +169,76 @@ var getCompletedCourses = async studentID => {
...
+		// Check for invalid formatting
+		if (!format.verifyUsername(username)) {
+			error = 'Invalid format username';
+		} else if (!format.verifyPassword(password)) {
+			error = 'Invalid format password';
+		}
+
+		if (error) {
+			return done(error);
+		}
```
Ty-Won: Nit: on line 204 and 206 you can just write return done(the messages you have written)

backend/logic/users/users.js
GarethDPeters: no reason, but all of our other functions are vars rather than lets. Was the only reason i did it

backend/app.js
```diff
@@ -12,6 +13,11 @@ let tagsRouter = require('./routes/tags');
 
 let app = express();
 
+// For Passport
+app.use(session({ secret: 'keyboard cat',resave: true, saveUninitialized:true})); // session secret
```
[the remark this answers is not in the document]
GarethDPeters: so this is only useful if we decide to use sessions. I removed the functionality from the actual logic functions since it's not necessary right now. I can comment out this line if you want

backend/logic/users/users.js
GarethDPeters: The function returns before the end of the try catch, so it would not actually hit the finally.

KevenLYF: not really, the wait `finally` works is to clean up, so even if you throw an error or return, the finally block will always be executed.  except when:

The thread running the try-catch-finally block is killed or interrupted
You use System.exit(0);
The underlying VM is destroyed in some other way
The underlying hardware is unusable in some way

backend/logic/users/users.js
Ty-Won: @GarethDPeters I'd say for future reference, use let please :)

PR #36, opened 2019-02-22
Added API for retrieving user completed courses, major and minor

zerfuth:
This is for retrieving "student user data" as per the assigned task. I'm not sure if this will be needed in the end.

### DIFF
--- a/backend/logic/users/users.js
+++ b/backend/logic/users/users.js
@@ -112,7 +112,7 @@ var insertAdminUser = async (username, password, email, id) => {
   let hash = hasher.hashPass(password);
   try {
     await connection.beginTransaction();
-    await connection.query("INSERT INTO users VALUES(?, ?, ?);", [
+    await connection.query("INSERT INTO users (username, email, password) VALUES(?, ?, ?);", [
       username,
       email,
       hash
--- a/backend/logic/users/users.js
+++ b/backend/logic/users/users.js
@@ -77,35 +77,31 @@ var getCompletedCourses = async studentID => {
 var getStudentData = async studentID => {
   let data;
 
-  let sql_query = `SELECT course_code, semester
-  		FROM course_offerings
-  		WHERE (id, semester)
-  		IN (SELECT offering_id, semester FROM student_course_offerings WHERE student_id = ?);`;
-
   let conn = await mysql.getNewConnection();
 
-  let courses;
+  let courses, major, minor;
   try {
-    courses = await conn.query(sql_query, [studentID]);
+    await conn.begintransaction();
+    courses = await conn.query(
+      `SELECT course_code, semester
+    FROM course_offerings
+    WHERE (id, semester)
+    IN (SELECT offering_id, semester FROM student_course_offerings WHERE student_id = ?);`,
+      [studentID]
+    );
+    major = await conn.query(
+      `SELECT curriculum_name FROM student_major WHERE student_id = ?;`,
+      [studentID]
+    );
+    minor = await conn.query(
+      `SELECT curriculum_name FROM student_minor WHERE student_id = ?;`,
+      [studentID]
+    );
+    conn.commit();
+    conn.release();
   } catch (err) {
-    console.log(err);
-  }
-
-  sql_query = `SELECT curriculum_name FROM student_major WHERE student_id = ?;`;
-
-  let major;
-  try {
-    major = await conn.query(sql_query, [studentID]);
-  } catch (err) {
-    console.log(err);
-  }
-
-  sql_query = `SELECT curriculum_name FROM student_minor WHERE student_id = ?;`;
-
-  let minor;
-  try {
-    minor = await conn.query(sql_query, [studentID]);
-  } catch (err) {
+    conn.rollback();
+    conn.release();
     console.log(err);
   }
 
--- a/backend/logic/users/users.js
+++ b/backend/logic/users/users.js
@@ -50,7 +50,7 @@ var insertStudentUser = async (username, password, email, id) => {
   }
 };
 
-var deleteStudentUser = async (username) => {
+var deleteStudentUser = async username => {
   if (!format.verifyUsername(username)) {
     let error = "invalid username";
     console.error(error);
@@ -67,14 +67,19 @@ var deleteStudentUser = async (username) => {
 
   try {
     await connection.beginTransaction();
-    let student_id = await connection.query('SELECT student_id FROM students WHERE username = ?;', username)
-    await connection.query('DELETE FROM students WHERE student_id = ?;', [student_id[0].student_id]);
-    await connection.query('DELETE FROM users WHERE username = ?;', [username]);
+    let student_id = await connection.query(
+      "SELECT student_id FROM students WHERE username = ?;",
+      username
+    );
+    await connection.query("DELETE FROM students WHERE student_id = ?;", [
+      student_id[0].student_id
+    ]);
+    await connection.query("DELETE FROM users WHERE username = ?;", [username]);
     await connection.commit();
     connection.release();
     return true;
   } catch (error) {
-    console.error(error)
+    console.error(error);
     connection.rollback();
     connection.release();
     throw new Error(false);
@@ -83,9 +88,9 @@ var deleteStudentUser = async (username) => {
 
 var insertAdminUser = async (username, password, email, id) => {
   // Connect to database
-  let error = false
+  let error = false;
   // Check for invalid formatting
-  //todo, handle errors after formatting configured to throw errors 
+  //todo, handle errors after formatting configured to throw errors
   if (!format.verifyUsername(username)) {
     error = "invalid format username";
   } else if (!format.verifyPassword(password)) {
@@ -107,8 +112,15 @@ var insertAdminUser = async (username, password, email, id) => {
   let hash = hasher.hashPass(password);
   try {
     await connection.beginTransaction();
-    await connection.query("INSERT INTO users VALUES(?, ?, ?);", [username, email, hash]);
-    await connection.query("INSERT INTO staff_members VALUES(?, ?);", [id, username]);
+    await connection.query("INSERT INTO users VALUES(?, ?, ?);", [
+      username,
+      email,
+      hash
+    ]);
+    await connection.query("INSERT INTO staff_members VALUES(?, ?);", [
+      id,
+      username
+    ]);
     await connection.commit();
     connection.release();
     return true;
@@ -120,7 +132,7 @@ var insertAdminUser = async (username, password, email, id) => {
   }
 };
 
-var deleteAdminUser = async (username) => {
+var deleteAdminUser = async username => {
   if (!format.verifyUsername(username)) {
     let error = "invalid username";
     console.error(error);
@@ -137,14 +149,19 @@ var deleteAdminUser = async (username) => {
 
   try {
     await connection.beginTransaction();
-    let staff_id = await connection.query('SELECT staff_id FROM staff_members WHERE username = ?;', username)
-    await connection.query('DELETE FROM staff_members WHERE staff_id = ?;', [staff_id[0].staff_id]);
-    await connection.query('DELETE FROM users WHERE username = ?;', [username]);
+    let staff_id = await connection.query(
+      "SELECT staff_id FROM staff_members WHERE username = ?;",
+      username
+    );
+    await connection.query("DELETE FROM staff_members WHERE staff_id = ?;", [
+      staff_id[0].staff_id
+    ]);
+    await connection.query("DELETE FROM users WHERE username = ?;", [username]);
     await connection.commit();
     connection.release();
     return true;
   } catch (error) {
-    console.error(error)
+    console.error(error);
     connection.rollback();
     connection.release();
     throw new Error(false);
@@ -158,7 +175,7 @@ var getCompletedCourses = async studentID => {
   		FROM course_offerings
   		WHERE (id, semester)
   		IN (SELECT offering_id, semester FROM student_course_offerings WHERE student_id = ?);`;
-  
+
   let connection = await mysql.getNewConnection();
   let results;
   try {
@@ -176,10 +193,18 @@ var getCompletedCourses = async studentID => {
 };
 
 var getStudentData = async studentID => {
+  let error = false;
+  if (!format.verifyStudentId(studentID)) {
+    error = "invalid format id";
+  }
+
+  if (!error == false) {
+    console.error(error);
+    throw new Error(error);
+  }
+
   let data;
-
   let conn = await mysql.getNewConnection();
-
   let courses, major, minor;
   try {
     courses = await conn.query(
--- a/backend/logic/users/users.js
+++ b/backend/logic/users/users.js
@@ -52,11 +52,11 @@ var getCompletedCourses = async studentID => {
   		WHERE (id, semester)
   		IN (SELECT offering_id, semester FROM student_course_offerings WHERE student_id = ?);`;
 
-  mysql.query = util.promisify(mysql.query);
+  let conn = await mysql.getNewConnection();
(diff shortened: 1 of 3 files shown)


backend/routes/users.js
```diff
@@ -77,6 +77,26 @@ router.get("/getCompletedCourses", async (req, res) => {
...
+ * @author: Feras Al Taha
+ *
+ */
+router.get("/getStudentData", async (req, res) => {
+  const student_id = req.query.studentID;
+	//const courses = await users.getCompletedCourses(student_id);
+	const data = await users.getStudentData(student_id);
+  res.status(200).send(data);
+});
+
```
AlexLam1997:
```suggestion
try {
const data = await users.getStudentData(student_id);
res.status(200).send(data);
}catch(error){
//for now just send back generic 500, will have to edit later on to return specific error codes based on what happened
res.status(500).send(error)
}
```

backend/routes/users.js
```diff
@@ -77,6 +77,26 @@ router.get("/getCompletedCourses", async (req, res) => {
...
+ * 		curl -X GET -H "Content-Type: application/json" 'http://localhost:3000/users/getStudentData?studentID=12345'
+ *
+ * @returns An json of current major & minors, and completed course code, e.g {"major":"Electrical Major","minor":"Software Minor","courses":[{"course_code":"ECSE422","semester":"W2019"},{"course_code":"ECSE428","semester":"F2019"}]}
+ *
+ * @author: Feras Al Taha
+ *
+ */
+router.get("/getStudentData", async (req, res) => {
+  const student_id = req.query.studentID;
+	//const courses = await users.getCompletedCourses(student_id);
```
AlexLam1997:
```suggestion
```

backend/logic/users/users.js
```diff
@@ -67,7 +67,51 @@ var getCompletedCourses = async studentID => {
...
+
+  let minor;
+  try {
+    minor = await conn.query(sql_query, [studentID]);
+  } catch (err) {
+    console.log(err);
+  }  
+
+  let results = {'major':major, 'minor':minor, 'courses':courses};
+
```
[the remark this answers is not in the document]
zerfuth: Makes sense! Updated the code to address this

backend/routes/users.js
```diff
@@ -77,6 +76,30 @@ router.get("/getCompletedCourses", async (req, res) => {
...
+  //const courses = await users.getCompletedCourses(student_id);
+  const data = await users.getStudentData(student_id);
+  res.status(200).send(data);
+});
+try {
+const data = await users.getStudentData(student_id);
+res.status(200).send(data);
+}catch(error){
+//for now just send back generic 500, will have to edit later on to return specific error codes based on what happened
+res.status(500).send(error)
```
si-x-li: Formatting is off here

backend/test/users_test.js
```diff
@@ -296,3 +296,18 @@ describe("Tests add student user", function() {
...
+describe("Test retrieve student data", function() {
+  it("responds with valid", function() {
+    return users.getStudentData(2).then(function(res) {
+      let found = false;
+      let searchingFor = {
+        major: [],
+        minor: [],
+        courses: [{ course_code: "ECSE 428", semester: "winter" }]
+      };
+      if (JSON.stringify(course) == JSON.stringify(searchingFor)) found = true;
```
si-x-li: As it's structured, the test is likely going to break as soon as we start adding more stuff into the database

backend/logic/users/users.js
```diff
@@ -67,7 +74,47 @@ var getCompletedCourses = async studentID => {
...
 };
 
+var getStudentData = async studentID => {
+  let data;
+
+  let conn = await mysql.getNewConnection();
+
+  let courses, major, minor;
+  try {
+    await conn.begintransaction();
```
si-x-li: No need to begin transaction since you're only dealing with read functions

backend/test/users_test.js
AlexLam1997: Try inserting random values into the necessary tables and then retrieving those in your test. Like this for example: 

it('responds with valid', async function(){
        connection = await mysql.getNewConnection()
        connection.query(`INSERT INTO courses (course_code,title, department) VALUES 
        (?,?,?) ON DUPLICATE KEY UPDATE course_code=course_code;`,['ECSE 428','Software engineering in practice', 'ECSE']);
        connection.query(`INSERT INTO tags(name) VALUES
        (?);`,['Engineering'])
        connection.query(`INSERT INTO course_tag (course_code, tag_name) VALUES
        (?,?);`,['ECSE 428', 'Engineering'])
        return courses.queryCourseByTag('Engineering').then(function(res){
            let found = false;
            let searchingFor = {course_code: 'ECSE 428'};
            for (course in res){
                if (JSON.stringify(course) == JSON.stringify(searchingFor))
                    found = true;
            }
            assert(true,found)
        });
    });
Make sure you insert random data in it so that such values are not already in the database and that you remove that random data afterwards (In my case ECSE 428 is pretty much a constant entry that should always be in the database so I did not bother, but it would have been better practice to insert some random value eg randomcourse123456) 
Also I'm not sure but shouldn't you be comparing JSON.stringify(res) == JSON.stringify(searchingFor) instead of course? what is course? :O

backend/routes/users.js
```diff
@@ -77,6 +76,30 @@ router.get("/getCompletedCourses", async (req, res) => {
...
+ * @apiExample {curl} Example usage:
+ * 		curl -X GET -H "Content-Type: application/json" 'http://localhost:3000/users/getStudentData?studentID=12345'
+ *
+ * @returns An json of current major & minors, and completed course code, e.g {"major":"Electrical Major","minor":"Software Minor","courses":[{"course_code":"ECSE422","semester":"W2019"},{"course_code":"ECSE428","semester":"F2019"}]}
+ *
+ * @author: Feras Al Taha
+ *
+ */
+router.get("/getStudentData", async (req, res) => {
+  const student_id = req.query.studentID;
```
AlexLam1997:
```suggestion
try {
const data = await users.getStudentData(student_id);
res.status(200).send(data);
}catch(error){
//for now just send back generic 500, will have to edit later on to return specific error codes based on what happened
res.status(500).send(error)
}
```

backend/routes/users.js
AlexLam1997: This block is not in the right place, see my above suggestion and please remove it

backend/routes/users.js
AlexLam1997: Remove the next 3 lines if taking this commit. The try catch block was not in the endpoint ;)

backend/test/users_test.js
Ty-Won: For easier reading of what he wrote @zerfuth 

```node
it("responds with valid", async function() {
  connection = await mysql.getNewConnection();
  connection.query(
    `INSERT INTO courses (course_code,title, department) VALUES (?,?,?) ON DUPLICATE KEY UPDATE course_code=course_code;`,
    ["ECSE 428", "Software engineering in practice", "ECSE"]
  );
  connection.query(`INSERT INTO tags(name) VALUES (?);`, ["Engineering"]);
  connection.query(
    `INSERT INTO course_tag (course_code, tag_name) VALUES (?,?);`,
    ["ECSE 428", "Engineering"]
  );
  return courses.queryCourseByTag("Engineering").then(function(res) {
    let found = false;
    let searchingFor = { course_code: "ECSE 428" };
    for (course in res) {
      if (JSON.stringify(course) == JSON.stringify(searchingFor)) found = true;
    }
    assert(true, found);
  });
});

```

backend/routes/users.js
zerfuth: gotcha, will fix

backend/logic/users/users.js
```diff
@@ -29,8 +30,15 @@ var insertStudentUser = async (username, password, email, id) => {
 
   try {
     await connection.beginTransaction();
-    await connection.query("INSERT INTO users VALUES(?, ?, ?);", [username, email, hash]);
-    await connection.query("INSERT INTO students VALUES(?, ?);", [id, username]);
+    await connection.query("INSERT INTO users VALUES(?, ?, ?);", [
```
AlexLam1997:
```suggestion
    await connection.query("INSERT INTO users (username, email, password) VALUES(?, ?, ?);", [
```

backend/logic/users/users.js
```diff
@@ -29,8 +30,15 @@ var insertStudentUser = async (username, password, email, id) => {
...
+      username,
+      email,
+      hash
+    ]);
+    await connection.query("INSERT INTO students VALUES(?, ?);", [
```
AlexLam1997:
```suggestion
    await connection.query("INSERT INTO students (student_id, username) VALUES(?, ?);", [
```

backend/logic/users/users.js
```diff
@@ -167,10 +175,47 @@ var getCompletedCourses = async studentID => {
...
+var getStudentData = async studentID => {
+  let data;
+
+  let conn = await mysql.getNewConnection();
+
+  let courses, major, minor;
+  try {
+    courses = await conn.query(
+      `SELECT course_code, semester
+    FROM course_offerings
```
AlexLam1997: picky, but formatting is off here

backend/logic/users/users.js
```diff
@@ -167,10 +175,47 @@ var getCompletedCourses = async studentID => {
...
+    conn.release();
+  } catch (err) {
+    conn.release();
+    console.log(err);
+  }
+
+  let results = { major: major, minor: minor, courses: courses };
+
+  if (results) {
+    data = JSON.stringify(results);
```
AlexLam1997: could you put this along with the return statement in the try catch too?

backend/logic/users/users.js
```diff
@@ -167,10 +175,47 @@ var getCompletedCourses = async studentID => {
...
+      `SELECT curriculum_name FROM student_major WHERE student_id = ?;`,
+      [studentID]
+    );
+    minor = await conn.query(
+      `SELECT curriculum_name FROM student_minor WHERE student_id = ?;`,
+      [studentID]
+    );
+    conn.release();
+  } catch (err) {
+    conn.release();
```
AlexLam1997: What do you return when an error occurs?

backend/routes/users.js
```diff
@@ -126,4 +126,27 @@ router.get("/getCompletedCourses", async (req, res) => {
...
+  try {
+    const data = await users.getStudentData(student_id);
+    res.status(200).send(data);
+  } catch (error) {
+    //for now just send back generic 500, will have to edit later on to return specific error codes based on what happened
+    res.status(500).send(error);
```
AlexLam1997:
```suggestion
    res.status(500).send(error.message);
```

AlexLam1997: Steven pointed out to me that the stack trace shouldnt be sent overto the client, so just send them the error message

backend/test/users_test.js
```diff
@@ -629,3 +630,40 @@ describe("Tests add admin user", function() {
...
+      `INSERT INTO courses (course_code,title, department) VALUES (?,?,?) ON DUPLICATE KEY UPDATE course_code=course_code;`,
+      ["ECSE 428", "Software engineering in practice", "ECSE"]
+    );
+    connection.query(
+      `INSERT INTO course_offerings (id, semester, course_code) VALUES (?,?,?) ON DUPLICATE KEY UPDATE id=id;`,
+      [321123, "winter", "ECSE 428"]
+    );
+    connection.query(
+      `INSERT INTO users (username,email, password) VALUES (?,?,?) ON DUPLICATE KEY UPDATE username=username;`,
+      ["testUser", "email@domain.com", 1234]
```
AlexLam1997: Check that this works with the new formatting Steven implemented, the constraints are more strict now. (Username length, password length to be checked)

AlexLam1997: I think min password length is 8 now

backend/test/users_test.js
```diff
@@ -629,3 +630,40 @@ describe("Tests add admin user", function() {
...
+    );
+    connection.query(
+      `INSERT INTO students (student_id,username) VALUES (?,?) ON DUPLICATE KEY UPDATE student_id=student_id;`,
+      [123321, "testUser"]
```
AlexLam1997: same, check that the formatting passes

backend/test/users_test.js
```diff
@@ -629,3 +630,40 @@ describe("Tests add admin user", function() {
...
+    );
+    connection.query(
+      `INSERT INTO student_course_offerings (student_id,offering_id,semester) VALUES (?,?,?);`,
+      [123321, 321123, "winter"]
```
AlexLam1997: Change student id accordingly here

backend/logic/users/users.js
zerfuth: Don't think it's necessary to have these if you fill in all columns. Also, didn't work on that method.

backend/logic/users/users.js
zerfuth: same as above

backend/logic/users/users.js
zerfuth: what part of it is off?

backend/logic/users/users.js
AlexLam1997: Its more prevention in case we add or remove a column from this table in the future, it would break your code

backend/logic/users/users.js
```diff
@@ -167,10 +175,47 @@ var getCompletedCourses = async studentID => {
...
+    console.log(err);
```
zerfuth:
```suggestion
    console.log(err);
    throw new Error(error);
```
Would this be better?

backend/logic/users/users.js
AlexLam1997: from should be indented more i think

backend/logic/users/users.js
zerfuth: ok makes sense

backend/logic/users/users.js
zerfuth: Not according to prettier plugin it seems. First line is indented but the subsequent ones aren't (like an indent at the start of a paragraph)

backend/logic/users/users.js
zerfuth: see suggestion below

backend/logic/users/users.js
AlexLam1997: Yes! Since its being caught at the endpoint level

backend/logic/users/users.js
AlexLam1997: oh.. ew

backend/logic/users/users.js
AlexLam1997: nice

backend/logic/users/users.js
```diff
@@ -99,8 +112,15 @@ var insertAdminUser = async (username, password, email, id) => {
   let hash = hasher.hashPass(password);
   try {
     await connection.beginTransaction();
-    await connection.query("INSERT INTO users VALUES(?, ?, ?);", [username, email, hash]);
-    await connection.query("INSERT INTO staff_members VALUES(?, ?);", [id, username]);
+    await connection.query("INSERT INTO users VALUES(?, ?, ?);", [
```
AlexLam1997: i know this isnt you but can you fix it here too please? 
```suggestion
    await connection.query("INSERT INTO users (username, email, password) VALUES(?, ?, ?);", [
```

backend/logic/users/users.js
```diff
@@ -99,8 +112,15 @@ var insertAdminUser = async (username, password, email, id) => {
...
+      username,
+      email,
+      hash
+    ]);
+    await connection.query("INSERT INTO staff_members VALUES(?, ?);", [
```
AlexLam1997:
```suggestion
    await connection.query("INSERT INTO staff_members (id, username) VALUES(?, ?);", [
```

PR #38, opened 2019-02-22
Alex/error handling

AlexLam1997:


backend/routes/users.js
```diff
@@ -71,12 +47,10 @@ router.post("/completedCourses/comparison", function(req, res) {
...
  */
-router.get("/getCompletedCourses", async (req, res) => {
+router.get("/getCompletedCourses", asyncMiddleware(async (req, res) => {
   const student_id = req.query.studentID;
   const courses = await users.getCompletedCourses(student_id);
-  res.status(200).send(courses);
-});
-
-
+  res.status(200).send(customResponse(courses, null));
```
KevenLYF: If you are using undefined in customReponse.js  shouldn't you neglect the error field here ?

KevenLYF:
Overall looks awesome, this is a really good example for handling errors using Express middleware, don't know if we can get bonus point of it :) . Now I think we should focus more on the features.

backend/app.js
```diff
@@ -1,23 +1,33 @@
...
 app.use('/courses', coursesRouter);
+app.use(function (err, req, res, next) {
+    console.error(err.stack)
+    if (!isNaN(err.code)) {
+        res.status(err.code)
+    } else {
+        res.status(500)
+    }
+    res.send(customResponse('An error occured', err))
+})
```
Ty-Won: What are you handing over here?

backend/routes/users.js
AlexLam1997: oh i didnt know that would work, ill fix and see what it gives me

backend/app.js
KevenLYF: It's an Express middleware to catch and handle errors. The errors are passed from next

AlexLam1997: The first argument is what goes in the "result" field of the response. Second is only if an error occurs

backend/test/users_test.js
```diff
@@ -190,52 +189,53 @@ describe("Tests add student user", async function () {
...
+  //       "I.Have.A.Really.Long.Email.Address@emailDomainToo.com",
+  //       "535235231"
+  //     )
+  //     .then(response => {
+  //       return new Promise(function (resolve) {
+  //         assert.equal(response, "true");
+  //         resolve();
+  //       }).then(done);
+  //     });
+  // });
```
si-x-li: No need to comment out tests. Mocha allows you to skip them using .skip

AlexLam1997: Oh interesting, ill do that instead thanks!

PR #39, opened 2019-02-23
Ability to add courses completed by the student

si-x-li:
* Reformatted Mathieu's branch to be more user friendly and compatible with database
* Added ability to add course offerings
* Added ability to fetch all courses
* Fixed up comments

### DIFF
--- a/backend/logic/courses/courses.js
+++ b/backend/logic/courses/courses.js
@@ -1,22 +1,25 @@
 const mysql = require("../../sql/connection");
-const format = require("../../validation/format")
+const format = require("../../validation/format");
 
 var queryCourseByTag = async function queryCourseByTag(tag) {
   if (!tag) {
-    throw Error("Undefined tag")
+    throw Error("Undefined tag");
   }
 
   let connection = await mysql.getNewConnection();
   try {
-    let courses = await connection.query("SELECT course_code FROM course_tag WHERE course_tag.tag_name LIKE ?", tag);
+    let courses = await connection.query(
+      "SELECT course_code FROM course_tag WHERE course_tag.tag_name LIKE ?",
+      tag
+    );
     connection.release();
-    return (JSON.parse(JSON.stringify(courses)));
+    return JSON.parse(JSON.stringify(courses));
   } catch (error) {
     connection.release();
     console.error(error);
     throw Error(error.message);
   }
-}
+};
 
 var addCompletedCourses = async (studentId, courses, section) => {
   if (!format.verifyStudentId(studentId)) {
@@ -35,15 +38,25 @@ var addCompletedCourses = async (studentId, courses, section) => {
     await connection.beginTransaction();
     // Obtain all current course offerings and put them into a hashtable for fast lookup
     let results = await connection.query("SELECT * FROM course_offerings;");
-    let hashTable = await populateHashTableCourseOfferings(results);
+    let hashTable = {};
+    for (let i = 0; i < results.length; i++) {
+      hashTable[
+        results[i].course_code + results[i].semester + results[i].section
+      ] = results[i].id;
+    }
 
-    // Insert each course of the student into the database. If a course does not exist then 
+    // Insert each course of the student into the database. If a course does not exist then
     // throw an error.
     for (let course in courses) {
       for (let i = 0; i < courses[course].length; i++) {
-        let id = hashTable[course + courses[course][i].semester + courses[course][i].section];
-        await connection.query("INSERT INTO student_course_offerings VALUES(?, ?, ?);",
-              [studentId, id, courses[course][i].semester]);
+        let id =
+          hashTable[
+            course + courses[course][i].semester + courses[course][i].section
+          ];
+        await connection.query(
+          "INSERT INTO student_course_offerings VALUES(?, ?, ?);",
+          [studentId, id, courses[course][i].semester]
+        );
       }
     }
 
@@ -59,14 +72,11 @@ var addCompletedCourses = async (studentId, courses, section) => {
 };
 
 async function populateHashTableCourseOfferings(results) {
-  let hashTable = {};
-  for (let i = 0; i < results.length; i++) {
-    hashTable[results[i].course_code + results[i].semester + results[i].section] = results[i].id;
-  }
+
   return hashTable;
 }
 
 module.exports = {
   queryCourseByTag,
   addCompletedCourses
-}
+};
--- a/backend/logic/courses/courses.js
+++ b/backend/logic/courses/courses.js
@@ -1,19 +1,72 @@
-const mysql = require('../../sql/connection');
+const mysql = require("../../sql/connection");
+const format = require("../../validation/format")
 
-async function queryCourseByTag(tag) {
-    if (!tag) {
-        throw Error("Undefined tag")
-    }
+var queryCourseByTag = async function queryCourseByTag(tag) {
+  if (!tag) {
+    throw Error("Undefined tag")
+  }
 
-    let connection = await mysql.getNewConnection();
-    try {
-        let courses = await connection.query('SELECT course_code FROM course_tag WHERE course_tag.tag_name LIKE ?', tag);
-        connection.release();
-        return (JSON.parse(JSON.stringify(courses)));
-    } catch (error) {
-        connection.release();
-        console.error(error);
-        throw Error(error.message);
-    }
+  let connection = await mysql.getNewConnection();
+  try {
+    let courses = await connection.query("SELECT course_code FROM course_tag WHERE course_tag.tag_name LIKE ?", tag);
+    connection.release();
+    return (JSON.parse(JSON.stringify(courses)));
+  } catch (error) {
+    connection.release();
+    console.error(error);
+    throw Error(error.message);
+  }
 }
-module.exports.queryCourseByTag = queryCourseByTag;
+
+var addCompletedCourses = async (studentId, courses, section) => {
+  if (!format.verifyStudentId(studentId)) {
+    throw new Error("invalid format id");
+  }
+  await format.verifyCompletedCourses(courses);
+
+  let connection;
+  try {
+    connection = await mysql.getNewConnection();
+  } catch (error) {
+    throw new Error("failed to establish connection with database");
+  }
+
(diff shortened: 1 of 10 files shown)


backend/logic/courses/courses.js
```diff
@@ -1,19 +1,72 @@
...
+  }
+
+  let connection = await mysql.getNewConnection();
+  try {
+    let courses = await connection.query("SELECT course_code FROM course_tag WHERE course_tag.tag_name LIKE ?", tag);
+    connection.release();
+    return (JSON.parse(JSON.stringify(courses)));
+  } catch (error) {
+    connection.release();
+    console.error(error);
```
KevenLYF: maybe it's not necessary to log the error now, since when you are throwing it, the message is already displayed in the output, and it might already in the log (not sure about this).

backend/logic/courses/courses.js
```diff
@@ -1,19 +1,72 @@
...
+    return true;
+  } catch (error) {
+    console.error(error);
+    await connection.rollback();
+    connection.release();
+    throw new Error(false);
+  }
+};
+
+async function populateHashTableCourseOfferings(results) {
```
KevenLYF: no need to use `async` since no `await` is being used in the function.

backend/routes/courses.js
```diff
@@ -3,24 +3,84 @@ const courses = require('../logic/courses/courses');
...
+    res.send(result);
+  } catch (error) {
+    console.log(error);
+    res.status(500).send(error.message);
+  }
+});
+
+/*
+* @api {post} /addCourses
+* @apiDescription This endpoint will add an admin and an associated user
```
AlexLam1997: Is this api info correct?

backend/logic/courses/courses.js
```diff
@@ -1,19 +1,72 @@
...
+    connection.release();
+    return (JSON.parse(JSON.stringify(courses)));
+  } catch (error) {
+    connection.release();
+    console.error(error);
+    throw Error(error.message);
+  }
+}
+
+var addCompletedCourses = async (studentId, courses, section) => {
```
AlexLam1997: why do we need the section to add a completed course? I dont believe it is being used in this function

backend/logic/courses/courses.js
si-x-li: I don't see the log if I don't do a console.error. Only the status of the endpoint call is shown

KevenLYF: ok nvm then

backend/logic/courses/courses.js
```diff
@@ -1,19 +1,72 @@
...
+  if (!format.verifyStudentId(studentId)) {
```
AlexLam1997: i believe the verifyStudentId method should throw the error itself, less code repeat that way

backend/logic/courses/courses.js
si-x-li: Fixed thanks!

backend/logic/courses/courses.js
```diff
@@ -1,19 +1,72 @@
...
+    throw new Error("invalid format id");
+  }
+  await format.verifyCompletedCourses(courses);
+
+  let connection;
+  try {
+    connection = await mysql.getNewConnection();
+  } catch (error) {
+    throw new Error("failed to establish connection with database");
```
AlexLam1997: Here the getNewConnection method will already throw an error "failed to connect to database if anything goes wrong, so i think you can let this get picked up by the top level error handler

backend/logic/courses/courses.js
```diff
@@ -1,19 +1,72 @@
...
-        connection.release();
-        return (JSON.parse(JSON.stringify(courses)));
-    } catch (error) {
-        connection.release();
-        console.error(error);
-        throw Error(error.message);
+  try {
+    await connection.beginTransaction();
+    // Obtain all current course offerings and put them into a hashtable for fast lookup
+    let results = await connection.query("SELECT * FROM course_offerings;");
```
AlexLam1997: I think you can just select the id, semester and section from this table instead of all columns

backend/logic/courses/courses.js
```diff
@@ -1,19 +1,72 @@
...
     }
+
+    await connection.commit();
+    connection.release();
+    return true;
+  } catch (error) {
+    console.error(error);
+    await connection.rollback();
+    connection.release();
+    throw new Error(false);
```
```suggestion
    throw new Error(error);
```

AlexLam1997: Please throw errors with detail on what went wrong for debugging purposes

backend/routes/courses.js
```diff
@@ -3,24 +3,84 @@ const courses = require('../logic/courses/courses');
...
+  let section = req.body.section;
+  let coursesJSON;
+  try {
+    coursesJSON = JSON.parse(req.body.courses);
+  } catch (error) {
+    res.status(500).send("invalid format courses");
+    return;
+  }
+
+  courses.addCompletedCourses(studentId, coursesJSON, section)
```
AlexLam1997: await this method and put it in the try catch block instead

backend/logic/courses/courses.js
si-x-li: Front end should never have access to any stack or database information due to security issues

backend/routes/courses.js
si-x-li: Done

backend/logic/courses/courses.js
si-x-li: Changed

backend/logic/courses/courses.js
si-x-li: Will be handled in a separate commit

backend/logic/courses/courses.js
si-x-li: Removed

backend/routes/courses.js
si-x-li: API information is updated to the newest now

backend/logic/courses/courses.js
si-x-li: changed thanks!

backend/logic/courses/courses.js
```diff
@@ -1,19 +1,153 @@
...
+ * @returns A list of courses in JSON format
+ * @throws Undefined tag if tag is null
+ *         error if MySQL connection failed
+ */
+var queryCourseByTag = async function queryCourseByTag(tag) {
+  if (!tag) {
+    throw Error("Undefined tag");
+  }
+
+  let connection = await mysql.getNewConnection();
```
Ty-Won: Should this be in the try statement?

si-x-li: Nope @AlexLam1997 changed the getNewConnection() so that it throws an error with the error message

backend/logic/courses/courses.js
```diff
@@ -1,19 +1,153 @@
...
+ * @param {JSON} courseOfferings
+ * @returns true if successful
+ * @throws error if MySQL connection failed
+ *         invalid format course offerings if JSON format is wrong
+ *         false if insertion failed
+ */
+var addCourseOfferings = async courseOfferings => {
+  await format.verifyCourseOffering(courseOfferings);
+  let connection = await mysql.getNewConnection();
+  let query = "INSERT INTO course_offerings VALUES (?, ?, ?, ?, ?);";
```
AlexLam1997: Dont you need to specify column names?

```suggestion
  let query = "INSERT INTO course_offerings (id, semester, scheduled_time, course_code, section) VALUES (?, ?, ?, ?, ?);";
```

backend/logic/courses/courses.js
```diff
@@ -1,19 +1,153 @@
...
+
+/**
+ * Adds a list of course offerings into the database
+ * @author Steven Li
+ * @param {JSON} courseOfferings
+ * @returns true if successful
+ * @throws error if MySQL connection failed
+ *         invalid format course offerings if JSON format is wrong
+ *         false if insertion failed
+ */
```
AlexLam1997: Should we also add the course associated with the course offering if it doesnt exit? Or have the user do that himself with the other endpoint

backend/logic/courses/courses.js
```suggestion
  let query = "INSERT INTO course_offerings (id, semester, scheduled_time, course_code, section) VALUES (?, ?, ?, ?, ?);";
```

backend/logic/tags/tags.js
```diff
@@ -0,0 +1,24 @@
+const mysql = require("../../sql/connection");
+
+var getAllTags = async () => {
+  let connection = await mysql.getNewConnection();
+  let tags;
+  let result = [];
+
+  try {
+    result = await connection.query("SELECT * FROM tags;");
```
AlexLam1997:
```suggestion
    result = await connection.query("SELECT name FROM tags;");
```

backend/logic/courses/courses.js
si-x-li: There's no need since there are exactly 5 columns in the database

Ty-Won: I think it will default, but yeah should specify columns in case we add a new column in

backend/logic/tags/tags.js
```diff
@@ -0,0 +1,24 @@
...
+    connection.release();
+  } catch (error) {
+    console.error(error);
+    connection.release();
```
Ty-Won: nit:can we write connection release here at the end instead of writing 2x

backend/validation/format.js
```diff
@@ -5,64 +5,168 @@ const MAX_EMAIL_LENGTH = 384;
...
+ *       "courses": {
+ *         "ECSE 428": [{"semester": "W2017", "section": 1}],
+ *         "ECSE 356": [{"semester": "S2019", "section": 2}],
+ *         "ECSE 422": [{"semester": "F2019", "section": 1},
+ *                      {"semester": "W2018", "section": 3}]
+ */
+var verifyCourse = async courses => {
+  let error = "invalid format courses";
+  for (let course in courses) {
+    console.log(course);
```
AlexLam1997:
```suggestion
```

backend/logic/courses/courses.js
si-x-li: Fixed

PR #42, opened 2019-02-23
1. fixed error handling of getCompletedCourses

KevenLYF:
2. removed useless dependencies

### DIFF
--- a/backend/logic/users/users.js
+++ b/backend/logic/users/users.js
@@ -29,14 +29,15 @@ var insertStudentUser = async (username, password, email, id) => {
 
   try {
     await connection.beginTransaction();
-    await connection.query(
-      "INSERT INTO users (username, email, password) VALUES(?, ?, ?);",
-      [username, email, hash]
-    );
-    await connection.query(
-      "INSERT INTO students (student_id, username) VALUES(?, ?);",
-      [id, username]
-    );
+    await connection.query("INSERT INTO users VALUES(?, ?, ?);", [
+      username,
+      email,
+      hash
+    ]);
+    await connection.query("INSERT INTO students VALUES(?, ?);", [
+      id,
+      username
+    ]);
     await connection.commit();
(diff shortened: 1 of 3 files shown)


backend/logic/users/users.js
```diff
@@ -150,21 +174,23 @@ var getCompletedCourses = async studentID => {
...
   let results;
   try {
-    results = await connection.query(sql_query, [studentID]);
-    connection.release();
+    results = await conn.query(sql_query, [studentID]);
+    if (results.length !== 0) {
+      courses = JSON.stringify(results);
+      return courses;
+    }
+
```
zerfuth:
```suggestion
connection.release();
```
Still need to release connection, no?

zerfuth: NVM didn't see finally clause

KevenLYF:
@zerfuth  thanks, yeah it will always be executed no matter what you are doing before

backend/logic/users/users.js
```diff
@@ -150,21 +174,23 @@ var getCompletedCourses = async studentID => {
...
+    return "No student ID found!";
```
AlexLam1997: Its also possible that the student simply doesnt have any completed courses, I would change this to initialise courses to be an empty array, update it if (results.length!=0) or else return empty array 
```
let courses = []
if (results.length !== 0) {
      courses = JSON.stringify(results);
    }
return courses;

```

KevenLYF:
@AlexLam1997  Actually I thought about it too, but the table `student_course_offering` is meant to record the courses that a student has completed, so it cannot be empty.

AlexLam1997:
> @AlexLam1997 Actually I thought about it too, but the table `student_course_offering` is meant to record the courses that a student has completed, so it cannot be empty.

Hm but what if the student is newly registered in our service and has not yet imputted his list of completed courses? Would it not be empty then?

PR #44, opened 2019-02-23
DropDown Component

fionahang:
I added the DropDown component that is used in the CourseRegistration page and the Curriculum page.


client/src/components/DropDown/DropDown.js
```diff
@@ -0,0 +1,46 @@
...
+import PropTypes from 'prop-types';
+import Select from '@material-ui/core/Select';
+import MenuItem from '@material-ui/core/MenuItem';
+import './DropDown.css';
+
+const overrideStyle = {
+  color: '#FFFFFF',
+  fontSize: '0.8rem',
+  padding: '3px 0 3px 15px',
+};
```
fionahang: I had to override the style inline for these properties, because it wouldn't work with just the className and stylesheet. If anyone has a better solution, feel free to change it, but this works for now!

PR #46, opened 2019-02-24
Backend create curriculum finished and endpoint defined

GarethDPeters:


backend/logic/curriculums/curriculums.js
```diff
@@ -0,0 +1,95 @@
+const mysql = require("../../sql/connection");
+const format = require("../../validation/format");
+
+var createCurriculum = async (name, type, department, numOfElectives, cores, techComps, comps) => {
+    let error = false;
```
KevenLYF: Is it 4 spaces here?

KevenLYF:
Format is a bit messy, other than that overall looks good

backend/logic/curriculums/curriculums.js
```diff
@@ -0,0 +1,95 @@
...
+    } else if(!format.verifyNumOfElectives(numOfElectives)) {
+      error = "Invalid number of electives"
+    }
+  
+    if (!error == false) {
+      console.error(error);
+      throw new Error(error);
+    }
+  
+    format.verifyCourse(cores);
```
zerfuth: Shouldn't you return an error if upon verification, it turns out to be invalid?

backend/routes/curriculums.js
```diff
@@ -0,0 +1,59 @@
+const express = require("express");
+const curriculums = require("../logic/curriculums/curriculums");
+const router = express.Router();
+
+/**
+ * @api {post} /createCurriculum
+ * @apiDescription This endpoint will add a student and an associated user
```
zerfuth: Update the api description and params

backend/routes/curriculums.js
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * @author: Gareth Peters
+ */
+router.post("/createCurriculum", function(req, res) {
+	const name = req.body.name;
+	const type = req.body.type;
+	const department = req.body.department;
+	const numOfElectives = req.body.numOfElectives;
+	const cores = req.body.core;
+    const techComps = req.body.techComps;
```
zerfuth: Inconsistent indentation?

backend/logic/curriculums/curriculums.js
```diff
@@ -0,0 +1,95 @@
...
+      
+      await connection.commit();
+      connection.release();
+      return true;
+    } catch (error) {
+      connection.rollback();
+      connection.release();
+      console.error(error);
+      throw new Error(false);
+    }
```
AlexLam1997: Could you put the connection.release() in a finally block instead? 
```
try{
    ..
}catch{
    ..
} finally{
    conection.release
}
```

backend/logic/curriculums/curriculums.js
Ty-Won: good verification, follow up to feras's comment, if it throws an error, how will you handle it? Could you potentially put it within the try catch below? This way you lockdown any errors that happen

backend/logic/curriculums/curriculums.js
```diff
@@ -0,0 +1,95 @@
...
+  
+    format.verifyCourse(cores);
+    format.verifyCourse(techComps);
+    format.verifyCourse(comps);
+  
+    let connection = await mysql.getNewConnection();
+  
+    try {
+      await connection.beginTransaction();
+      await connection.query("INSERT INTO curriculums (curriculum_name, type, department, numOfElectives) VALUES(?, ?, ?, ?);", [
```
Ty-Won: Just to confirm is this to create a curriculum for the user? If it is just to create a curriculum for a year for a program, there is a script that was made (just needs to be fine tuned and merged)

backend/validation/format.js
```diff
@@ -86,6 +88,50 @@ var verifyAdminId = id => {
...
 
+/**
+ * Verifies that the curriculum name only contains alphanumerical characters
+ * @param {string} curriculum
+ */
+var verifyCurriculumName = curriculum => {
+  if (!curriculum || String(curriculum).length > MAX_CURR_NAME_LENGTH) {
+    return false;
+  }
+  return isAlphanumeric(username);
```
Ty-Won: My fault, you couldnt have known so not your fault, I wrote some curriculums with like a - in between words., just a heads up, is there a way to include the - symbol ?

GarethDPeters: I'll look into it

backend/logic/curriculums/curriculums.js
GarethDPeters: This function is called by routes, where the function is surrounded by a try catch and ends back a 500 if an error was detected.

backend/logic/curriculums/curriculums.js
GarethDPeters: The thing is you can't really do that, because if the an error is thrown and the program goes into the catch, the connection will never actually be released since an error is thrown. In other words, if it hits the catch, it will never go into the finally.

KevenLYF: nope, the finally will always run no matter what happened before

backend/logic/curriculums/curriculums.js
GarethDPeters: It was to create a particular curriculum for a year which was to be done by admins. But we could also use it for creating a curriculum for a user for sure

backend/logic/curriculums/curriculums.js
GarethDPeters: kk ill fix it then

backend/logic/curriculums/curriculums.js
zerfuth: Ok sorry, didn't realize that verifyCourses() threw an error and didnt just return a boolean like the other format validation functions.

PR #49, opened 2019-02-24
Remove course (phase out a course) 

AlexLam1997:


backend/logic/courses/courses.js
```diff
@@ -145,9 +147,26 @@ var addCourseOfferings = async courseOfferings => {
   }
 };
```
zerfuth: Please specify an API description for the function
```suggestion
/**
 * Phases out a course that is no longer offered
 * @author Alex Lam
 * @param {string} courseCode
 * @returns true if successful
 * @throws error if MySQL connection failed
 *         invalid format course code if course code format is incorrect
 *         false if modification failed
*/
```

backend/logic/courses/courses.js
```diff
@@ -145,9 +147,26 @@ var addCourseOfferings = async courseOfferings => {
...
 
+let phaseOutCourse = async courseCode => {
```
zerfuth: Make sure you do a validation check for the course code.

backend/logic/courses/courses.js
AlexLam1997: I put it with the endpoint in the other file

backend/logic/courses/courses.js
AlexLam1997: true, thanks!

backend/logic/courses/courses.js
zerfuth: yeah but from i thought we were doing an API description for the endpoint and also one for the actual methods?

backend/logic/courses/courses.js
```diff
@@ -145,9 +147,29 @@ var addCourseOfferings = async courseOfferings => {
...
+  let connection = await mysql.getNewConnection();
+  try {
+    if (!courseCode) {
```
zerfuth: check format of course code? format.js might already have a function for that, otherwise could write a simple one

Ty-Won:
great job, would it be possible to change the pull request title? It's simply for commit history reasons so that when we look back if we need to reset, we at least know what this is all about

backend/logic/courses/courses.js
Ty-Won: Try a regex expression, this is just a quick one but it should match 4 capital letters a single space and 3 digits 0-9
use https://regex101.com/ to help you if you do regex
```regex
courseCode.match(/[A-Z]{4}[ ]{1}\d{3}/g)
```

backend/logic/courses/courses.js
```diff
@@ -145,9 +147,27 @@ var addCourseOfferings = async courseOfferings => {
   }
 };
 
+let phaseOutCourse = async courseCode => {
+  let connection = await mysql.getNewConnection();
+  try {
+    await format.verifyCourse(courseCode);
+    await connection.query(
+      "UPDATE courses SET phased_out = TRUE WHERE course_code = ?",
```
Ty-Won: Not 100% sure, but from stevens explanation from the slack 

> @channel I don't think I explained myself clear during our meeting for beginning a transaction in MySQL. There is absolutely no need to begin and commit a transaction unless if you're writing data into the database (UPDATE, DELETE, INSERT, etc.). By default, MySQL database have auto commits turned on, but, in some cases, you want to temporarily disable auto commits until you're sure a block of queries will execute successfully which is where transactions come in. A transaction is used to preserve atomicity on multiple queries meaning that if one of the queries fails then the entire transaction must be rollback 

TLDR; If you're working with SELECTs only, you don't need to begin a transaction since it prevents legitimate write instructions from accessing the table

Do you have to use the transaction and commit function?

backend/logic/courses/courses.js
zerfuth: also, i think someone made a function for that (at least i saw it on @GarethDPeters's branch, so you could just reuse that.

backend/logic/courses/courses.js
si-x-li: A race condition is possible where connection A removes data before connection B updates it but this a likely extremely rare race condition considering how often courses get phased out (maybe once every 5 years) and either way if the call fails there's no net impact on the database so it's more for styling points lol

backend/logic/courses/courses.js
si-x-li: You can use **verifyCourseCode()** which just got merged into dev

backend/logic/courses/courses.js
AlexLam1997: whaat, its the same description though thats so redundant, I dont think we should do that

backend/logic/courses/courses.js
AlexLam1997: done! thanks :D

backend/logic/courses/courses.js
zerfuth: The route might be calling more than one function, each function should have a description of what it does. Just like javadocs man
At least that's my opinion on it, it's minor thing.

backend/logic/courses/courses.js
AlexLam1997: done!

db/update scripts/1.03.sql
```diff
@@ -0,0 +1,6 @@
+SET SQL_SAFE_UPDATES
```
si-x-li: Not a fan of turning off SQL_SAFE_UPDATES. Is there anyway to proceed without having this?

backend/logic/courses/courses.js
```diff
@@ -261,12 +263,30 @@ var updateCourse = async (course, newTitle, newTags) => {
...
+      courseCode
+    );
+    return true;
+  } catch (error) {
+    console.error(error);
+    throw error;
```
si-x-li: Not recommended to throw back SQL errors back at frontend since users will be able to see the structure of the database. Consider returning a custom message or masking the error instead.

backend/logic/courses/courses.js
```diff
@@ -261,12 +263,30 @@ var updateCourse = async (course, newTitle, newTags) => {
   }
 };
 
+let phaseOutCourse = async courseCode => {
+  let connection = await mysql.getNewConnection();
+  try {
+    await format.verifyCourse(courseCode);
```
si-x-li: This can be moved outside of the try block since you're rethrowing the error anyways

db/update scripts/1.03.sql
AlexLam1997: Not really since I need to update all the entries where phased out is currently null :/ No way to write it in terms of the course_code

backend/logic/courses/courses.js
AlexLam1997: Youre right ill fix that thanks!

backend/logic/courses/courses.js
AlexLam1997: Ahh I see what you mean. Alright ill add it thanks

PR #50, opened 2019-02-24
Gareth/backend/login wo passport

GarethDPeters:


backend/routes/users.js
```diff
@@ -131,6 +131,19 @@ router.get("/getCompletedCourses", async (req, res) => {
   return res.status(200).send(result);
 });
 
+/* Retrieve user's username and password and compare to stored values for logging in */
+router.post("/login", async (req, res) => {
+  const username = req.body.username;
+  const password = req.body.password;
+  try {
+    if (users.login(username, password)) {
+      res.status(200).send("Authenticated");
```
si-x-li: Just a question... Since users.login is asynchronous, won't Node be completely oblivious and proceed on running further codes? In this case, since there's no more code, then the request will resolve to nothing?

si-x-li: Just a question... Since users.login is asynchronous, won't Node be completely oblivious and proceed on running further codes? In this case, since there's no more code, then the request will resolve to nothing?

PR #52, opened 2019-02-24
Backend Feature (Story 5):

KevenLYF:
Assign a course to a curriculum

### DIFF
--- a/backend/logic/courses/courses.js
+++ b/backend/logic/courses/courses.js
@@ -289,6 +289,97 @@ let phaseOutCourse = async courseCode => {
   }
 };
 
+/**
+ * Assign a course to a curriculum
+ * @author Yufei Liu
+ * @param {string} courseType,{string} courseCode,{string} curriculum
+ * @returns true if successful
+ * @throws error if courseType, courseCode or curriculum dose not exist
+ *         course already assigned a curriculum
+ *
+ */
+var assignCourseToCurriculum = async (courseType, courseCode, curriculum) => {
+  if (!format.isMcGillCourse(courseCode)) {
+    throw Error("Invalid course format!\n");
+  }
+
+  let conn = await mysql.getNewConnection();
+
+  const checkExistQuery = `SELECT COUNT(*) AS count
+  FROM (
+      SELECT course_code, curriculum_name FROM curriculum_tech_comps
+      union all
+      SELECT course_code, curriculum_name FROM curriculum_complementaries
+      union all
+      SELECT course_code, curriculum_name FROM curriculum_core_classes
+  ) a
+  WHERE course_code = ? AND curriculum_name = ?;`;
+
+  // Check if the course has already been assigned, verify if curriculum type, course and curriculum are existed in DB
+  let ifAssigned;
+  let checkCourse;
+  let checkCurriculum;
+
+  try {
+    ifAssigned = await conn.query(checkExistQuery, [courseCode, curriculum]);
+    checkCourse = await conn.query(
+      "SELECT COUNT(*) AS count FROM courses WHERE course_code = ?",
+      [courseCode]
+    );
+    checkCurriculum = await conn.query(
+      "SELECT COUNT(*) AS count FROM curriculums WHERE curriculum_name = ?",
+      [curriculum]
+    );
+  } catch (err) {
+    console.log(err);
+    conn.release();
+    throw Error("Internal Server Error!\n");
+  }
+
+  if (ifAssigned[0].count !== 0) {
+    throw Error(
+      `Course ${courseCode} has already been added to ${curriculum}!\n`
+    );
+  }
+  if (checkCourse[0].count === 0) {
+    throw Error(`Course ${courseCode} does not exist!\n`);
+  }
+  if (checkCurriculum[0].count === 0) {
+    throw Error(`Curriculum ${curriculum} does not exist!\n`);
+  }
+
+  const coreTable = "curriculum_core_classes";
+  const techCompTable = "curriculum_tech_comps";
+  const compTable = "curriculum_complementaries";
+  let tableType = "";
+
+  if (courseType === "core") {
+    tableType = coreTable;
+  } else if (courseType === "techComp") {
+    tableType = techCompTable;
+  } else if (courseType === "complementaries") {
+    tableType = compTable;
+  } else {
+    throw Error("Invalid curriculum type!\n");
+  }
+
+  const query = `INSERT INTO  ${tableType} (curriculum_name, course_code)
+                VALUE (?, ?);`;
+
+  try {
+    await conn.beginTransaction();
+    await conn.query(query, [curriculum, courseCode]);
+
+    await conn.commit();
+  } catch (err) {
+    await conn.rollback();
+    console.log(err);
+    throw Error("Internal Server Error!\n");
+  } finally {
+    conn.release();
+  }
+};
+
 module.exports = {
   queryCourseByTag,
   addCompletedCourses,
@@ -297,5 +388,6 @@ module.exports = {
   addCoreq,
   addPrereq,
   updateCourse,
-  phaseOutCourse
+  phaseOutCourse,
+  assignCourseToCurriculum
 };
--- a/backend/routes/courses.js
+++ b/backend/routes/courses.js
@@ -334,4 +334,41 @@ router.post("/phaseOutCourse", async (req, res, next) => {
   }
 });
 
+/**
+ *
+ * @api {post} /assignCourseToCurriculum
+ * @apiDescription assign a course to a courseType (core, techComp, complementaries)
+ * @apiParam (body) {string} courseType, {string} courseCode, {string} curriculum
+ * @apiExample {curl} Example usage:
+ *	curl -X POST \
+ *  -H 'Content-Type: application/json' \
+ *  -d '{"courseType": "core", "courseCode": "ECSE 428", "curriculum": "ee"}' \
+ *  http://localhost:3001/courses/assignCourseToCurriculum
+ *
+ * @returns True on success
+ *          invalid courseType
+ *          curriculum not exists
+ *          course not exists
+ *          course already assigned to selected curriculumn
+ *
+ * @author: Yufei Liu
+ *
+ */
+router.post("/assignCourseToCurriculum", async (req, res) => {
+  const courseType = req.body.courseType;
+  const courseCode = req.body.courseCode;
+  const curriculum = req.body.curriculum;
+
+  try {
+    await courses.assignCourseToCurriculum(courseType, courseCode, curriculum);
+    res.status(200).send(true);
+  } catch (err) {
+    if (err.message === "Internal Server Error!\n") {
+      res.status(500).send(err.message);
+    } else {
+      res.status(400).send(err.message);
+    }
+  }
+});
+
 module.exports = router;
--- a/backend/validation/format.js
+++ b/backend/validation/format.js
@@ -292,5 +292,6 @@ module.exports = {
   verifyCourseOffering,
   verifyCoreq,
   verifyPrereq,
-  verifyCourseCode
+  verifyCourseCode,
+  isMcGillCourse
 };


backend/logic/courses/courses.js
```diff
@@ -261,12 +261,102 @@ var updateCourse = async (course, newTitle, newTags) => {
...
+ *
+ */
+var assignCourseToCurriculum = async (
+  curriculumType,
+  courseCode,
+  curriculum
+) => {
+  let conn = await mysql.getNewConnection();
+
+  const checkExistQuery = `SELECT *
```
si-x-li: I may be wrong, but this prevents a course from being assigned to multiple curriculum.  For example, if a course_A is assigned to curriculum_A, it will be returned and you're no longer able to assign it to curriculum_B

KevenLYF: Ohhh you are right, I forget to  check curriculum. It should check duplicate for both course_code and curriculum.

AlexLam1997: I agree I think you should check if there are duplicate course_codes within the given curriculum only :)

backend/logic/courses/courses.js
```diff
@@ -261,12 +261,102 @@ var updateCourse = async (course, newTitle, newTags) => {
...
+
+  // Check if the course has already been assigned, verify if curriculum type, course and curriculum are existed in DB
+  let ifAssigned;
+  let checkCourse;
+  let checkCurriculum;
+
+  try {
+    ifAssigned = await conn.query(checkExistQuery, [courseCode]);
+    checkCourse = await conn.query(
+      "SELECT * FROM courses WHERE course_code = ?",
```
```suggestion
      "SELECT COUNT(*) FROM courses WHERE course_code = ?",
```

AlexLam1997: Since you are only using this to check if the course exists you dont need to select * which grabs all the columns in the table. You just need the count of rows that match your criteria

backend/logic/courses/courses.js
```diff
@@ -261,12 +261,104 @@ var updateCourse = async (course, newTitle, newTags) => {
...
+  let checkCurriculum;
+
+  try {
+    ifAssigned = await conn.query(checkExistQuery, [courseCode, curriculum]);
+    checkCourse = await conn.query(
+      "SELECT * FROM courses WHERE course_code = ?",
+      [courseCode]
+    );
+    checkCurriculum = await conn.query(
+      "SELECT * FROM curriculums WHERE curriculum_name = ?",
```
```suggestion
      "SELECT COUNT(*) FROM curriculums WHERE curriculum_name = ?",
```

backend/logic/courses/courses.js
si-x-li: This would be good for now. For the future, we can add a composite key on course_code and curriculum to prevent insertion

backend/logic/courses/courses.js
```diff
@@ -261,12 +261,104 @@ var updateCourse = async (course, newTitle, newTags) => {
...
+    );
+  } catch (err) {
+    console.log(err);
+    conn.release();
+    throw Error("Internal Server Error!\n");
+  }
+
+  if (ifAssigned.length !== 0) {
+    throw Error(
+      `Course ${courseCode} has already been assigned with a curriculum!\n`
```
AlexLam1997: I think the message here should be that the course has already been added to the selected curriculum since a course is technically able to be assigned to multiple curriculums

backend/logic/courses/courses.js
```diff
@@ -261,12 +261,104 @@ var updateCourse = async (course, newTitle, newTags) => {
...
+    );
+  }
+  if (checkCourse.length === 0) {
```
AlexLam1997: If you use my count suggestion above this would become 
```
```
```suggestion
  if (checkCourse == 0 ){
```

backend/logic/courses/courses.js
```diff
@@ -261,12 +261,104 @@ var updateCourse = async (course, newTitle, newTags) => {
...
+ *         course already assigned a curriculum
+ *
+ */
+var assignCourseToCurriculum = async (
+  curriculumType,
+  courseCode,
+  curriculum
+) => {
+  let conn = await mysql.getNewConnection();
+
```
si-x-li: Maybe check if courseCode is in a valid format before proceeding with codes below

backend/logic/courses/courses.js
```diff
@@ -261,12 +261,104 @@ var updateCourse = async (course, newTitle, newTags) => {
...
+  if (ifAssigned.length !== 0) {
+    throw Error(
+      `Course ${courseCode} has already been assigned with a curriculum!\n`
+    );
+  }
+  if (checkCourse.length === 0) {
+    throw Error(`Course ${courseCode} does not exist!\n`);
+  }
+  if (checkCurriculum.length === 0) {
```
```suggestion
  if (checkCurriculum ==  0) {
```

AlexLam1997: same as above

backend/routes/courses.js
```diff
@@ -298,4 +298,45 @@ router.post("/updateCourse", async (req, res, next) => {
   }
 });
 
+/**
+ *
+ * @api {post} /assignCourseToCurriculum
+ * @apiDescription assign a course to a curriculum (core, techComp, complementaries)
+ * @apiParam (body) {string} curriculumType, {string} courseCode, {string} curriculum
```
AlexLam1997: Can you specify what values the curriculumType param should take? since the type has to have exact values to match

backend/routes/courses.js
```diff
@@ -298,4 +298,45 @@ router.post("/updateCourse", async (req, res, next) => {
...
+ * @apiExample {curl} Example usage:
+ *	curl -X POST \
+ *  -H 'Content-Type: application/json' \
+ *  -d '{"curriculumType": "core", "courseCode": "ECSE 428", "curriculum": "ee"}' \
+ *  http://localhost:3001/courses/assignCourseToCurriculum
+ *
+ * @returns None
```
AlexLam1997:
```suggestion
 * @returns True on success
```

backend/routes/courses.js
```diff
@@ -298,4 +298,45 @@ router.post("/updateCourse", async (req, res, next) => {
...
 
+/**
+ *
+ * @api {post} /assignCourseToCurriculum
+ * @apiDescription assign a course to a curriculum (core, techComp, complementaries)
+ * @apiParam (body) {string} curriculumType, {string} courseCode, {string} curriculum
+ * @apiExample {curl} Example usage:
+ *	curl -X POST \
+ *  -H 'Content-Type: application/json' \
+ *  -d '{"curriculumType": "core", "courseCode": "ECSE 428", "curriculum": "ee"}' \
```
si-x-li: It might be beneficial in terms of overall speed to process a course with a batch of curriculums for a single class instead of one at a time. Maybe something like this

```
{
  "course": "ECSE 428",
  "curriculums": [
    {"type": "core", "curriculum": "ee"},
    {"type": "core", "curriculum": "se-minor"}
  ]
}
```

This would need significant change and would be better left for next sprint

KevenLYF: I thought about it actually, but I'm not sure what the front-end would be like.  For example if they just using buttons to assign a course to one curriculum and submit, what I have now is sufficient.  I suggest to keep it like this, and if what you suggested is needed by the front-end, we definitely can improve it.

backend/routes/courses.js
KevenLYF: Actually I put it in line 304, gonna update curriculum to curriculumType

backend/logic/courses/courses.js
KevenLYF: The reason why I didn't use COUNT(*) is because it will return `RowDataPacket { 'COUNT(*)': 0 }`, then I need to convert it to a JSON, and read the `0` in the JSON, which would be too much. So I prefer to keep it like this, what do you think ?

backend/logic/courses/courses.js
KevenLYF: follow up my reply above, `if (checkCourse === 0 )` won't work since it returns a data package

backend/logic/courses/courses.js
AlexLam1997: Hm yeah that thats an interesting problem. I had to try out a few things before I figure it out. In the query you can actually just do 
```
SELECT COUNT(*) AS count 
..
```
And then access it through 
```
result[0].count
```

backend/logic/courses/courses.js
AlexLam1997: Check my reply above on accessing the datapacket

backend/routes/courses.js
AlexLam1997: Ah ok i see, thats good then!

backend/logic/courses/courses.js
KevenLYF: thanks! gonna fix it now

backend/routes/courses.js
AlexLam1997: Since we're pressed for time I agree to leave it like this for now as well. Should be modified to what steven suggested in the future though

KevenLYF: That's definitely a good improvement and I'm open to it. However, I would suggest to hear from front-end first and see how they are designing the UI to assign a course.
If they are assigning a course to a list of curriculum, we could go with what Steven said, but if they decided to assign it one by one, then we can keep this.

backend/logic/courses/courses.js
```diff
@@ -261,12 +261,108 @@ var updateCourse = async (course, newTitle, newTags) => {
   }
 };
 
+/**
+ * Assign a course to a curriculum
+ * @author Yufei Liu
+ * @param {string} curriculumType,{string} courseCode,{string} curriculum
```
zerfuth: curriculumType should really be courseType because at first I thought you were referring to major vs minor or even 7 sem vs 8 sem curriculum.

backend/logic/courses/courses.js
```diff
@@ -261,12 +261,108 @@ var updateCourse = async (course, newTitle, newTags) => {
...
+  let conn = await mysql.getNewConnection();
+
+  const checkExistQuery = `SELECT COUNT(*) AS count
+  FROM (
+      SELECT course_code, curriculum_name FROM curriculum_tech_comps
+      union all
+      SELECT course_code, curriculum_name FROM curriculum_complementaries
+      union all
+      SELECT course_code, curriculum_name FROM curriculum_core_classes
+  ) a
```
zerfuth: is the a after the parenthesis a typo? i don't recognize such syntax and i didn't find anything online that corresponds to this

backend/logic/courses/courses.js
```diff
@@ -261,12 +261,108 @@ var updateCourse = async (course, newTitle, newTags) => {
...
+    tableType = compTable;
+  } else {
+    throw Error("Invalid curriculum type!\n");
+  }
+
+  const query = `INSERT INTO  ${tableType} (curriculum_name, course_code)
+                VALUE (?, ?);`;
+
+  try {
+    await conn.beginTransaction();
```
zerfuth: Someone with better SQL experience can probably address this better than me but since you have a single SQL query to execute, wouldn't the transaction be unnecessary? It's not like you have multiple queries that need to be run sequentially and depend on each other.

backend/routes/courses.js
```diff
@@ -298,4 +298,45 @@ router.post("/updateCourse", async (req, res, next) => {
...
+ * @apiDescription assign a course to a curriculumType (core, techComp, complementaries)
+ * @apiParam (body) {string} curriculumType, {string} courseCode, {string} curriculum
+ * @apiExample {curl} Example usage:
+ *	curl -X POST \
+ *  -H 'Content-Type: application/json' \
+ *  -d '{"curriculumType": "core", "courseCode": "ECSE 428", "curriculum": "ee"}' \
+ *  http://localhost:3001/courses/assignCourseToCurriculum
+ *
+ * @returns True on success
+ *          curriculumn curriculumType not existes
```
zerfuth:
```suggestion
 *          curriculum curriculumType not exists
```
Typo on this line and the three following lines (exists* instead of existes and curriculum* instead of curriculumn)

backend/logic/courses/courses.js
KevenLYF: make sense, will change that

backend/logic/courses/courses.js
KevenLYF: No it's not a typo, it's an alias, I think it's needed when you have nested statements. @si-x-li Correct me if I'm wrong

backend/logic/courses/courses.js
KevenLYF: I see your point, however the transaction here is more like a lock to avoid race condition, since I'm doing `write` operation here.

backend/logic/courses/courses.js
si-x-li: It's equivalent to saying AS a

zerfuth: didn't know about that!

backend/logic/courses/courses.js
zerfuth: Ok, not totally convinced that it is necessary but it's fine the way it is. Planning to discuss this further with @si-x-li in person.

PR #53, opened 2019-02-24
Yueh/client/main page

UAce:
I only display information. I did not implement any write functionalities to save the data.
The complementary section is wrong because the courses gotten from the tags are main courses..


KevenLYF:
Consider `amend`  next time so you won't have that much commits history :)

PR #54, opened 2019-02-25
Mat/backend/createCourse

MatSavoie:


backend/logic/courses/courses.js
```diff
@@ -29,6 +29,43 @@ var queryCourseByTag = async function queryCourseByTag(tag) {
...
+ * @author Mathieu Savoie
+ * @param {String} courseCode
+ * @param {String} title
+ * @param {String} departement
+ * @returns true if successful
+ * @throws error if MySQL connection failed
+ *         invalid format courses if JSON format is incorrect
+ *         false if insertion failed
+ */
+var addCourse = async (courseCode, title, departement, phasedOut) => {
```
AlexLam1997: The default in the database for phasedOut is false so you dont explicitally have to specify it when inserting. Because of this I think it would be better if the front end was not forced to send a "phased out" param when adding a course into the db 
```suggestion
var addCourse = async (courseCode, title, departement) => {
```

backend/logic/courses/courses.js
```diff
@@ -29,6 +29,43 @@ var queryCourseByTag = async function queryCourseByTag(tag) {
...
+  
+   try {
+    await connection.beginTransaction();
+    await connection.query("INSERT INTO courses VALUES(?, ?, ?, ?);", [courseCode, title, departement, phasedOut]);
+    await connection.commit();
+    connection.release();
+    return true;
+  } catch (error) {
+    connection.rollback();
+    connection.release();
```
si-x-li: connection.release() can be put in the finally block as recommended by @KevenLYF

backend/validation/format.js
```diff
@@ -275,6 +288,36 @@ var verifyCourseCode = async courseCode => {
...
 
+/**
+ * Verifies that the title is not null
+ * @param {String} title
+ */
+var verifyTitle = async title => {
+  if (!title){
+		throw new Error("invalid format title");
+	}
+};
```
si-x-li: Formatting needs to be fixed

backend/logic/courses/courses.js
```diff
@@ -29,6 +29,43 @@ var queryCourseByTag = async function queryCourseByTag(tag) {
...
+ *         false if insertion failed
+ */
+var addCourse = async (courseCode, title, departement, phasedOut) => {
+  
+  
+  // Verifying proper format
+  await format.verifyCourseCode(courseCode);
+  await format.verifyTitle(title);
+  await format.verifyDepartmentSubName(departement);
+  await format.verifyPhaseOut(phasedOut);
```
AlexLam1997: Same thing here, should be able to create a course in the db without having to hand in a phased out param (also to be corrected in the format.verifyPhasedOut method

backend/logic/courses/courses.js
```diff
@@ -29,6 +29,43 @@ var queryCourseByTag = async function queryCourseByTag(tag) {
...
+  
+  // Connect to database
+  let connection = await mysql.getNewConnection();
+  
+   try {
+    await connection.beginTransaction();
+    await connection.query("INSERT INTO courses VALUES(?, ?, ?, ?);", [courseCode, title, departement, phasedOut]);
```
```suggestion
    await connection.query("INSERT INTO courses VALUES(?, ?, ?);", [courseCode, title, departement]);
```

backend/logic/courses/courses.js
```diff
@@ -29,6 +29,43 @@ var queryCourseByTag = async function queryCourseByTag(tag) {
...
+    await connection.commit();
+    connection.release();
```
AlexLam1997: do connection.release() in a finally block after try and catch

backend/routes/courses.js
```diff
@@ -20,6 +20,48 @@ router.get("/getCourseByTag", async function(req, res, next) {
...
+		"departement" : "ECSE"
+	}
+* Curl:
+	curl -X POST \
+	http://localhost:3001/courses/createCourse\
+	-H 'Content-Type: application/json' \
+	-d '{
+		"courseCode": "ECSE 428",
+		"title": "Software Engineering Practice",
+		"departement" : "ECSE"
```
si-x-li: Update your comments to indicate phased_out as part of the body

backend/routes/courses.js
```diff
@@ -20,6 +20,48 @@ router.get("/getCourseByTag", async function(req, res, next) {
...
+*
+* @returns true if course was added successfully or false otherwise
+*
+* @author: Mathieu Savoie
+*/
+router.post('/createCourse', function(req, res, next) {
+  const courseCode = req.body.courseCode;
+  const title = req.body.title;
+  const departement = req.body.departement;
+  const phasedOut = req.body.phasedOut
```
AlexLam1997: you have phased out here, but not in the params specified in the comment above. Careful about this. Works out well in this case though since we should remove it from params anyways (or put it as a optional param by allowing it to be indefined, up to you )

backend/validation/format.js
```diff
@@ -275,6 +288,36 @@ var verifyCourseCode = async courseCode => {
   }
 };
 
+/**
+ * Verifies that the title is not null
+ * @param {String} title
+ */
+var verifyTitle = async title => {
+  if (!title){
+		throw new Error("invalid format title");
```
AlexLam1997: fix formatting here

backend/validation/format.js
```diff
@@ -275,6 +288,36 @@ var verifyCourseCode = async courseCode => {
...
+		throw new Error("invalid format department name");
+	}
+};
+
+/**
+ * Verifies that the phased_out is either 0 or 1
+ * @param {String} phaseOut
+ */
+var verifyPhaseOut = async phaseOut => {
+	if (!phaseOut || !(String(phaseOut).match(/^[01]$/i)) ){
```
AlexLam1997: Not sure what the formatting is checking here, phased out should be a boolean

si-x-li: He assumes that either 0 (false) or 1 (true) is transmitted from front end

AlexLam1997: does casting a boolean to a string return a digit (0/1) or "false"/"true" if its the later then this wont work

backend/logic/courses/courses.js
```diff
@@ -29,6 +29,43 @@ var queryCourseByTag = async function queryCourseByTag(tag) {
...
+  await format.verifyCourseCode(courseCode);
+  await format.verifyTitle(title);
+  await format.verifyDepartmentSubName(departement);
+  await format.verifyPhaseOut(phasedOut);
+  
+  // Connect to database
+  let connection = await mysql.getNewConnection();
+  
+   try {
+    await connection.beginTransaction();
```
zerfuth:
```suggestion
```
Unnecessary transaction since you are executing a single query. Use transactions when you have multiple queries that modify the db or depend on each other.

backend/logic/courses/courses.js
```diff
@@ -29,6 +29,43 @@ var queryCourseByTag = async function queryCourseByTag(tag) {
...
+    await connection.query("INSERT INTO courses VALUES(?, ?, ?, ?);", [courseCode, title, departement, phasedOut]);
+    await connection.commit();
```
zerfuth:
```suggestion
```
See comment above

backend/routes/courses.js
```diff
@@ -20,6 +20,48 @@ router.get("/getCourseByTag", async function(req, res, next) {
   }
 });
 
+/*
+* @api {post} /createCourse
+* @apiDescription This endpoint will add a course and an associated user
```
zerfuth:
```suggestion
* @apiDescription This endpoint will add a course
```
Unless I misunderstood your feature, I believe you are only adding a course without associating any user.

backend/logic/courses/courses.js
MatSavoie: I tried following your instructions and looking online, but i get a bunch of errors... I have an assignment due in 1 hour that I have to finish, I can work on this during the break if necessary!

AlexLam1997: no problem this is great!

PR #68, opened 2019-03-06
fixed a bunch of errors found when testing the code, refactored endpoint responses, tested all endpoints

AlexLam1997:


backend/logic/courses/courses.js
```diff
@@ -373,9 +344,8 @@ var assignCourseToCurriculum = async (courseType, courseCode, curriculum) => {
       [curriculum]
     );
   } catch (err) {
-    console.log(err);
     conn.release();
-    throw Error("Internal Server Error!\n");
```
KevenLYF: I'm using this string to classify 500 (Internal error) , how do you know if it's an internal error if remove it? I checked the `customeResponse` didn't see any 500 handler, did I miss something or ?

backend/logic/courses/courses.js
```diff
@@ -38,37 +34,30 @@ var queryCourseByTag = async function queryCourseByTag(tag) {
...
-    ]);
-    return true;
-  } catch (error) {
-    connection.rollback();
-    console.error(error);
-    throw new Error(false);
-  } finally {
-    connection.release();
-  }
+  await connection.query("INSERT INTO courses VALUES(?, ?, ?, ?);", [
```
si-x-li: Add column fields to the query

backend/logic/users/users.js
```diff
@@ -29,157 +17,128 @@ var insertStudentUser = async (username, password, email, id) => {
 
   try {
     await connection.beginTransaction();
-    await connection.query("INSERT INTO users VALUES(?, ?, ?);", [username, email, hash]);
-    await connection.query("INSERT INTO students VALUES(?, ?);", [id, username]);
+    await connection.query("INSERT INTO users VALUES(?, ?, ?);", [
```
si-x-li: Add column values to remain consistent with other queries

backend/logic/users/users.js
```diff
@@ -29,157 +17,128 @@ var insertStudentUser = async (username, password, email, id) => {
...
+      username,
+      email,
+      hash
+    ]);
+    await connection.query("INSERT INTO students VALUES(?, ?);", [
```
si-x-li: Same as above

backend/logic/users/users.js
```diff
@@ -29,157 +17,128 @@ var insertStudentUser = async (username, password, email, id) => {
...
 
   let connection = await mysql.getNewConnection();
 
   // Hash the password
   let hash = hasher.hashPass(password);
   try {
     await connection.beginTransaction();
-    await connection.query("INSERT INTO users VALUES(?, ?, ?);", [username, email, hash]);
-    await connection.query("INSERT INTO staff_members VALUES(?, ?);", [id, username]);
+    await connection.query("INSERT INTO users VALUES(?, ?, ?);", [
```
si-x-li: Same here

backend/logic/users/users.js
```diff
@@ -29,157 +17,128 @@ var insertStudentUser = async (username, password, email, id) => {
...
+      username,
+      email,
+      hash
+    ]);
+    await connection.query("INSERT INTO staff_members VALUES(?, ?);", [
```
si-x-li: Same here

backend/logic/users/users.js
```diff
@@ -29,157 +17,128 @@ var insertStudentUser = async (username, password, email, id) => {
...
 
   try {
     await connection.beginTransaction();
-    let staff_id = await connection.query("SELECT staff_id FROM staff_members WHERE username = ?;", username);
-    await connection.query("DELETE FROM staff_members WHERE staff_id = ?;", [staff_id[0].staff_id]);
+    let staff_id = await connection.query(
+      "SELECT staff_id FROM staff_members WHERE username = ?;",
+      username
+    );
+    await connection.query("DELETE FROM staff_members WHERE staff_id = ?;", [
```
si-x-li: Pretty sure we're able to reduce these 2 queries into a single one without causing any issues

DELETE FROM staff_members WHERE staff_id IN (SELECT staff_id FROM staff_members WHERE username = ?);

backend/logic/courses/courses.js
AlexLam1997: It defaults to sending back a 500 error code if no error code is specified. If you want to specify a specific code you can do so with: 
```
let error = new Error ("message") 
error.code = 400; 
throw error; 
```

PR #71, opened 2019-03-11
Feras/backend/get student data

zerfuth:
Endpoint for retrieving a student's major, minors, completed courses, missing core classes offered next year and desired tech comps offered next year.


backend/test/users_test.js
```diff
@@ -623,40 +623,41 @@ describe("Tests add admin user", function() {
...
+//         major: [],
+//         minor: [],
+//         courses: [{ course_code: "ECSE 428", semester: "winter" }]
+//       };
+//       if (JSON.stringify(res) == JSON.stringify(searchingFor)) {
+//         found = true;
+//       }
+//       assert(true, found);
+//     });
+//   });
```
KevenLYF: are there any operations for removing the test data inserted ?

backend/logic/users/users.js
```diff
@@ -170,9 +194,9 @@ var getCompletedCourses = async studentID => {
...
   let error = false;
-  if (!format.verifyStudentId(studentID)) {
-    error = "invalid format id";
-  }
+  // if (!format.verifyStudentId(studentID)) {
+  //   error = "invalid format id";
+  // }
 
   if (!error == false) {
     console.error(error);
```
Ty-Won: With the way this is written now, the if statement will never run

backend/logic/users/users.js
```diff
@@ -181,20 +205,117 @@ var getStudentData = async studentID => {
...
+      AND (semester = ? OR semester = ?) 
+      AND (course_code 
+          IN (SELECT course_code 
+              FROM curriculum_tech_comps 
+              WHERE curriculum_name = ?)) 
+      AND (course_code
+          IN (SELECT course_code 
+              FROM student_desired_courses 
+              WHERE student_id = ?));`,
+      [studentID, fallSem, winterSem, curriculumName, studentID]
```
Ty-Won: explain the query to me to help me understand it

backend/test/users_test.js
zerfuth: No this test is outdated hence why I commented it out. Whoever takes the task for testing this endpoint will have to rewrite most of it.

backend/logic/users/users.js
zerfuth: Select all courses that are not in the set of completed courses AND that are in the correct semester (offered next year) AND that are in the set of valid techcomps for the major AND that are in the set of desired courses to take

backend/logic/users/users.js
```diff
@@ -181,20 +205,117 @@ var getStudentData = async studentID => {
...
+      `SELECT curriculum_name FROM student_minors WHERE student_id = ?;`,
+      [studentID]
+    );
+
+    let currYear = new Date().getFullYear();
+    let currMonth = new Date().getMonth();
+    let fallSem, winterSem;
+    let curriculumName = major[0].curriculum_name;
+
+    if (currMonth < 3) {
```
si-x-li: What about Summer?

backend/logic/users/users.js
si-x-li: Is there a way to reduce the number of tables being queried? Or can you pull data from the tables you need and put them in a hashtable? As more data gets inserted in the database, these queries are going to become way too slow

zerfuth: Reduced it by one. I know it's still not optimal but I suggest we just make a low priority task for further optimizing if that's fine given there are more urgent tasks for now.

backend/logic/users/users.js
zerfuth: Current webscraping doesn't include summer courses so it should be fine but I agree that it should be supported. I suggest adding a low priority task for refactoring this code to include summer terms and minors later on :)

Ty-Won:
@zerfuth resolved conflicts, please confirm these changes I've made are as desired. If so merge

zerfuth:
seems good to me

backend/logic/users/users.js
```diff
@@ -194,9 +194,9 @@ var getCompletedCourses = async studentID => {
...
 var getStudentData = async studentID => {
   let error = false;
-  // if (!format.verifyStudentId(studentID)) {
-  //   error = "invalid format id";
-  // }
+  if (!format.verifyStudentId(studentID)) {
+    error = "invalid format id";
+  }
 
   if (!error == false) {
```
KevenLYF: You can simply do 
```
let error;
...
if (error) {
// handler
}
```

backend/logic/users/users.js
```diff
@@ -181,20 +181,111 @@ var getStudentData = async studentID => {
...
+              WHERE curriculum_name = ?)) 
+      AND (SELECT course_code 
+          FROM student_desired_courses 
+          WHERE (student_id = ?, desired = TRUE));`,
+      [studentID, fallSem, winterSem, curriculumName, studentID]);
+    
+    // For each incomplete course, get prereqs and coreqs
+    // for(course c in courses){
+    //   c[prereqs] = [];
+    //   c[coreqs] = [];
```
KevenLYF: you can remove the comments here

backend/logic/users/users.js
zerfuth: true true

PR #74, opened 2019-03-13
Adding and removing tags

si-x-li:
Adding abilities to create, remove tags and assign tags to courses

### DIFF
--- a/backend/logic/tags/tags.js
+++ b/backend/logic/tags/tags.js
@@ -35,9 +35,8 @@ var getAllTags = async () => {
  *         invalid format tag if tag is null
  */
 var addTag = async tag => {
+  format.verifyTag(tag);
   let connection = await mysql.getNewConnection();
-  await format.verifyTag(tag);
-
   try {
     await connection.query("INSERT INTO tags (name) VALUES (?);", tag);
     return true;
@@ -59,9 +58,8 @@ var addTag = async tag => {
  *         invalid format tag if tag is null
  */
 var removeTag = async tag => {
+  format.verifyTag(tag);
   let connection = await mysql.getNewConnection();
-  await format.verifyTag(tag);
-
   try {
     await connection.beginTransaction();
     await connection.query("DELETE FROM course_tags WHERE tag_name = ?;", tag);
@@ -83,14 +81,14 @@ var removeTag = async tag => {
  * @param {Array} tags An array of tags
  */
 var assignTagsToCourse = async (course, tags) => {
-  let connection = await mysql.getNewConnection();
-  await format.verifyCourseCode(course);
+  format.verifyCourseCode(course);
   if (!tags) {
     throw new Error("invalid format tag");
   }
   for (let i = 0; i < tags.length; i++) {
-    await format.verifyTag(tags[i]);
+    format.verifyTag(tags[i]);
   }
+  let connection = await mysql.getNewConnection();
 
   try {
     await connection.beginTransaction();
--- a/backend/logic/tags/tags.js
+++ b/backend/logic/tags/tags.js
@@ -1,5 +1,12 @@
 const mysql = require("../../sql/connection");
+const format = require("../../validation/format");
 
+/**
+ * Returns all tags
+ * @author Steven Li
+ * @returns A list of tags in JSON format
+ * @throws error if MySQL connection failed
+ */
 var getAllTags = async () => {
   let connection = await mysql.getNewConnection();
   let tags;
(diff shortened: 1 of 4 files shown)


backend/routes/tags.js
```diff
@@ -18,4 +18,109 @@ router.get("/getAllTags", async function(req, res, next) {
...
+ * Curl:
+ *	curl -X POST
+ *	http://localhost:3001/tags/deleteTag
+ *	-H 'Content-Type: application/json'
+ *	-d '{
+ *       "tag": "Engineering"
+ *      }'
+ * @author: Steven Li
+ */
+router.post("/deleteTag", async function (req, res, next) {
```
Ty-Won: Could you use the delete rest request?

backend/logic/tags/tags.js
```diff
@@ -18,6 +25,102 @@ var getAllTags = async () => {
...
+ * Adds a tag into the database.
+ * @author Steven Li
+ * @param {String} tag
+ * @returns true if successful
+ * @throws error if MySQL connection failed
+ *         false if insertion failed
+ *         invalid format tag if tag is null
+ */
+var addTag = async tag => {
+  let connection = await mysql.getNewConnection();
```
zerfuth: Minor detail but I think you should get the connection after verifying the tag format

backend/logic/tags/tags.js
```diff
@@ -18,6 +25,102 @@ var getAllTags = async () => {
...
+ * Removes a tag from the database.
+ * @author Steven Li
+ * @param {String} tag
+ * @returns true if successful
+ * @throws error if MySQL connection failed
+ *         false if deletion failed
+ *         invalid format tag if tag is null
+ */
+var removeTag = async tag => {
+  let connection = await mysql.getNewConnection();
```
zerfuth: same as above

backend/logic/tags/tags.js
```diff
@@ -18,6 +25,102 @@ var getAllTags = async () => {
...
+  }
+};
+
+/**
+ * Assigns tags to a course.
+ * @param {String} course A course code
+ * @param {Array} tags An array of tags
+ */
+var assignTagsToCourse = async (course, tags) => {
+  let connection = await mysql.getNewConnection();
```
zerfuth: same as above

backend/routes/tags.js
si-x-li: Fixed

backend/logic/tags/tags.js
si-x-li: Good catch would have caused problems otherwise

backend/logic/tags/tags.js
```diff
@@ -18,6 +25,102 @@ var getAllTags = async () => {
...
+  await format.verifyCourseCode(course);
+  if (!tags) {
+    throw new Error("invalid format tag");
+  }
+  for (let i = 0; i < tags.length; i++) {
+    await format.verifyTag(tags[i]);
```
KevenLYF: no need to use wait, I think verifyTag is not an async operation

backend/logic/tags/tags.js
```diff
@@ -18,6 +25,102 @@ var getAllTags = async () => {
...
+  try {
+    await connection.beginTransaction();
+    // Remove old entries
+    await connection.query(
+      "DELETE FROM course_tags WHERE course_code = ?;",
+      course
+    );
+
+    // Add new entries
+    for (let i = 0; i < tags.length; i++) {
```
KevenLYF: just a minor suggestion, you can use functional programming.
e.g
```
tags. forEach(tag => {
// do sth
})
```

backend/logic/tags/tags.js
KevenLYF: minor suggestion, you can also do 
```
tags.forEach( tag => {
// do stuff
})

File: backend/routes/tags.js [hunk 7]
```diff
@@ -18,4 +18,109 @@ router.get("/getAllTags", async function(req, res, next) {
...
+ * @author: Steven Li
+ */
+router.post("/assignTagsToCourse", async function (req, res, next) {
+  let course = req.body.course;
+  let tag;
+
+  try {
+    tag = JSON.parse(req.body.tags);
+  } catch (err) {
+    res.status(500).send("invalid format tags");
```
KevenLYF: 404

backend/test/tags_test.js
```diff
@@ -0,0 +1,79 @@
...
+  });
+
+  it("responds with invalid format tag on delete", async () => {
+    await tags.removeTag(null).catch(response => {
+      assert.equal(response.message, expectedInvalidFormat);
+    });
+  });
+
+  it("responds with true", async () => {
+    before(async () => {
```
KevenLYF: question, why do you put `before` and `after` in a `it` ?

backend/validation/format.js
```diff
@@ -324,6 +324,16 @@ var verifyPhaseOut = phaseOut => {
   }
 };
 
+/**
+ * Verifies that the tag is not null
+ * @param {String} tag
+ */
+var verifyTag = async tag => {
```
KevenLYF: don't think this is an async function

backend/logic/tags/tags.js
```diff
@@ -18,6 +25,102 @@ var getAllTags = async () => {
...
+/**
+ * Assigns tags to a course.
+ * @param {String} course A course code
+ * @param {Array} tags An array of tags
+ */
+var assignTagsToCourse = async (course, tags) => {
+  let connection = await mysql.getNewConnection();
+  await format.verifyCourseCode(course);
+  if (!tags) {
+    throw new Error("invalid format tag");
```
KevenLYF: `Invalid format tags array` ?

PR #76, opened 2019-03-17
Tyrone/backend/db/add course details

Ty-Won:
Added curriculum description and credits


Ty-Won:
Resolving an issue with the queries

PR #78, opened 2019-03-17
Updated getCourseByTag to also return description and whether it is d…

zerfuth:
…esired by student

### DIFF
--- a/backend/logic/courses/courses.js
+++ b/backend/logic/courses/courses.js
@@ -4,15 +4,13 @@ const format = require("../../validation/format");
 /**
  * Returns a list of courses matching the tag
  * @author Alex Lam, Feras Al Taha
- * @param {string} tag
+ * @param {string} tag, studentID
  * @returns A list of courses in JSON format
  * @throws Undefined tag if tag is null
  *         error if MySQL connection failed
  */
 var getCourseByTag = async function getCourseByTag(tag,studentID) {
-  if (!tag) {
-    throw Error("Undefined tag");
-  }
+  format.verifyTag(tag);
   format.verifyStudentId(studentID);
 
   let connection = await mysql.getNewConnection();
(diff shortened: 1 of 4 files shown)


backend/logic/courses/courses.js
```diff
@@ -3,23 +3,31 @@ const format = require("../../validation/format");
 
 /**
  * Returns a list of courses matching the tag
- * @author Alex Lam
+ * @author Alex Lam, Feras Al Taha
  * @param {string} tag
```
AlexLam1997: Update params please

backend/logic/courses/courses.js
```diff
@@ -3,23 +3,31 @@ const format = require("../../validation/format");
...
  * @returns A list of courses in JSON format
  * @throws Undefined tag if tag is null
  *         error if MySQL connection failed
  */
-var getCourseByTag = async function getCourseByTag(tag) {
+var getCourseByTag = async function getCourseByTag(tag,studentID) {
   if (!tag) {
```
AlexLam1997: you can use format.verifyTag here, it does the same

AlexLam1997:
Also you have to fix the tests associated with get course by tag, theyre broken now because of this change

backend/logic/courses/courses.js
```diff
@@ -3,23 +3,31 @@ const format = require("../../validation/format");
...
  * Returns a list of courses matching the tag
- * @author Alex Lam
+ * @author Alex Lam, Feras Al Taha
  * @param {string} tag
  * @returns A list of courses in JSON format
  * @throws Undefined tag if tag is null
  *         error if MySQL connection failed
  */
-var getCourseByTag = async function getCourseByTag(tag) {
+var getCourseByTag = async function getCourseByTag(tag,studentID) {
```
si-x-li: Isn't this endpoint used to return courses matching a tag for everyone and not for a specific student?

zerfuth: Well the page displaying the courses by tag also needed to show if those courses were marked as desired or not so we decided to modify this endpoint so that it would still return the same courses as before but with an additional parameter specifying if they were previously marked as desired or not by the student

PR #83, opened 2019-03-17
Tyrone/backend/db/add course details

Ty-Won:
Added course descriptions and credits


si-x-li:
Ran fine when I used it on Sunday

PR #84, opened 2019-03-17
Modified user/login endpoint, now it returns studentID on authenticated

KevenLYF:


backend/logic/users/users.js
```diff
@@ -72,15 +72,14 @@ var insertAdminUser = async (username, password, email, id) => {
...
-      username
-    ]);
+    await connection.query(
+      "INSERT INTO users (username, email, password) VALUES(?, ?, ?);",
+      [username, email, hash]
+    );
+    await connection.query(
+      "INSERT INTO staff_members (staff_id, username) VALUES(?, ?);",
+      [id, username]
+    );
```
KevenLYF: Ignore this part, just auto-formatting. No changes involved here

christiaua:
Maybe also return the name if it's not too hard to change, like that we don't need another endpoint to display the name in the header bar. Task: `Display the user's name in the Header Bar`

Ty-Won:
@KevenLYF will you be adding the name as @christiaua suggested? If so I will wait on that before merging

KevenLYF:
@christiaua  @Ty-Won  Sure that's a good idea, sry I've been too busy didn't see the reply. I'll try to finish that by tonight.

KevenLYF:
@christiaua  now it returns JSON `{"student_id":279263971,"username":"studentUser1"}`

christiaua:
After the meeting discussion, return the student's `email`

PR #85, opened 2019-03-17
Mathieu/backend/future course end point

MatSavoie:


backend/test/course_test.js
```diff
@@ -151,3 +152,94 @@ describe("Test assign course to curriculumn", () => {
...
+    await users.insertStudentUser (username, password, email, id);
+
+  });
+
+  it("responds with invalid format student id 1", function(done) {
+    courses
+      .saveUserPreferences(null, ["ECSE 307", "ECSE 251", "ECSE 325"])
+      .catch(response => {
+        return new Promise(function(resolve) {
+          assert.equal(response.message, "Id cannot be empty");
```
Ty-Won: Nice job matt,quick question why do you return a promise?

backend/test/course_test.js
```diff
@@ -151,3 +152,94 @@ describe("Test assign course to curriculumn", () => {
...
+      });
+  });
+
+  
+  it("responds with invalid format student id 2", function(done) {
+    courses
+      .saveUserPreferences(invalid_id_1, ["ECSE 307", "ECSE 251", "ECSE 325"])
+      .catch(response => {
+        return new Promise(function(resolve) {
+          assert.equal(response.message, "false");
```
Ty-Won: I know the format does not match but I think you should be testing response.message vs the actual id. Unless saveUserPreserences returns false, then if so ignore this

backend/logic/courses/courses.js
```diff
@@ -401,6 +401,50 @@ var assignCourseToCurriculum = async (courseType, courseCode, curriculum) => {
...
+ * @throws invalid format course code
+ *         invalid format student id
+ *         mysql connection errors
+ */
+var saveUserPreferences = async (student_id, courses) => {
+  if (!format.verifyStudentId(student_id)) {
+    throw new Error("invalid format student id");
+  }
+  if (!courses) {
+    throw new Error("invalid format course code");
```
zerfuth: Rather than "invalid format course code", wouldn't that be instead "empty courses list" or "no courses provided"?

backend/routes/courses.js
```diff
@@ -363,4 +363,52 @@ router.post(
...
+ *     }'
+ *
+ * @returns true if future desired courses were added successfully
+ *          false if future desired courses were not added
+ *          invalid format future courses if the courses format is invalid
+ *          database connection handling
+ *
+ * @author: Mathieu Savoie
+ */
+router.post("/updateFutureCourse", async (req, res, next) => {
```
zerfuth: I'd call this /updateDesiredCourse rather than future course because I think it's more correct. That said, it's just a suggestion so feel free to ignore that, it's a minor detail.

backend/logic/courses/courses.js
MatSavoie: Fixed! and added an extra test case for it

backend/routes/courses.js
MatSavoie: done

backend/test/course_test.js
MatSavoie: yeah it returns false in this case

backend/test/course_test.js
MatSavoie: followed legacy code implementation, I just made changes to them

PR #87, opened 2019-03-17
endpoint implemented

AlexLam1997:


backend/routes/curriculums.js
```diff
@@ -89,4 +89,12 @@ router.post(
   })
 );
 
+router.get(
+  "/getCurriculumYears",
+  asyncMiddleware(async function(req, res, next) {
```
Ty-Won: Just a question, does asyncMiddleware need async? is it redundant? Or do you need to specify async?

Ty-Won:
@zerfuth @AlexLam1997 to add in to this discussion, which tasks in the backlog is this covering? it makes sense to fetch the year whenever you retrieve curriculum.But in this case, if there's a certain scenario where you need to fetch all years only, then this is fine.

PR #90, opened 2019-03-19
Login functionality

christiaua:
- Send request if input is valid
- Added curriculum length to sign up


client/src/components/LoginForm/SignupContent.jsx
```diff
@@ -3,20 +3,18 @@ import Section from '../Section';
...
+const tempCurriculum = ['7 semesters', '8 semesters'];
+const tempMajors = ['Computer', 'Electrical', 'Software'];
 
 class SignupContent extends Component {
   render() {
     return (
       <Section className="form" flexDirection="column">
+        <text className="error" id="error-msg" />
         <Section className="subform" flexDirection="row">
           <EditText label="Full Name" /> &nbsp;
```
UAce: <EditText> component needs an **onChange** function as **props**
If you dont wanna implement it yet, you can just write a dummy function that does nothing

client/src/components/LoginForm/LoginContent.jsx
```diff
@@ -1,19 +1,120 @@
...
+    if (this.isInputValid()) {
+      //Send Request
+      axios
+        .post(LOGIN_URL, {
+          username: this.state.username,
+          password: this.state.password,
+        })
+        .then(response => {
+          // Got good response
+          console.log(response);
```
AlexLam1997: I think youd have to do a response.data.Response to get the actual "Authenticated" message we're sending back to you because of the new structure.. oops

client/src/components/LoginForm/LoginContent.jsx
```diff
@@ -1,19 +1,120 @@
...
+          // TODO: change page, save student ID and name when backend is done
+          // temp for testing
+          this.setError('login successful');
```
AlexLam1997: Confusion lol

client/src/components/LoginForm/LoginContent.jsx
```diff
@@ -1,19 +1,120 @@
...
+        })
+        .catch(loginError => {
+          // Invalid login
+          console.log(loginError);
+          var error = loginError.response.data.ErrorMessage;
+          if (
+            error === 'Password length must be between 8 and 64' ||
```
AlexLam1997: Theres also: 
"Incorrect username or password."
"Username length must be less than 64"
That would all kind of count as invalid credentials. The numbers are subject to change though (like we can suddenly just decide to put the max username length to 20 for example). So it may be simpler to just display the full error message instead but up to you.

client/src/components/LoginForm/LoginContent.jsx
```diff
@@ -1,19 +1,120 @@
...
+    // Check if username & password are empty
+    if (this.isEmpty(id)) {
+      this.setError('Student ID is required.');
+      return false;
+    } else if (this.isEmpty(password)) {
+      this.setError('Password is required.');
+      return false;
+    } else if (!/^\d+$/.test(id)) {
+      // Invalid Student ID (contains letters)
+      this.setInvalidCredentials();
```
AlexLam1997: Would it be better to display which field is causing the input to be invalid?

client/src/components/LoginForm/LoginContent.jsx
```diff
@@ -1,19 +1,120 @@
...
+  }
+
+  isEmpty(value) {
+    return !value;
+  }
+
+  handleInputChange(event = {}) {
+    const name = event.target && event.target.name;
+    const value = event.target && event.target.value;
+    this.setState({ [name]: value });
```
AlexLam1997: I'm confused as to what this means/does? are name and value booleans?

client/src/components/LoginForm/SignupContent.jsx
```diff
@@ -3,20 +3,18 @@ import Section from '../Section';
...
 
-const years = ['2014', '2015', '2016', '2017', '2018', '2019'];
-const majors = [
-  'Computer Engineering',
-  'Electrical Engineering',
-  'Software Engineering',
-];
+const tempYears = ['2014', '2015', '2016', '2017', '2018', '2019'];
+const tempCurriculum = ['7 semesters', '8 semesters'];
+const tempMajors = ['Computer', 'Electrical', 'Software'];
```
AlexLam1997: The names would have to be exactly
'Computer Engineering',
  'Electrical Engineering',
  'Software Engineering' if you want to hardcode it, but it would probably be better to grab that at the same time as the years now that I think of it. 
The curriculum would also have to be exactly "7-semester-curriculum" or "8-semester-curriculum". But also, should be grabbed from backend. I can change the endpoint for you.

client/src/components/LoginForm/LoginContent.jsx
christiaua: getting the email now, since it got pushed👌

client/src/components/LoginForm/LoginContent.jsx
christiaua: was for testing purposes 🤪

client/src/components/LoginForm/LoginContent.jsx
christiaua: The errors will be displayed in `Sign up`, for `Log in` it prevents people from attempting to guess credentials.

client/src/components/LoginForm/SignupContent.jsx
christiaua: I'll change the fields when sending it in the request, but we don't want to display `7-semester-curriculum` for the user. They don't need the dashed ☺️

client/src/components/LoginForm/LoginContent.jsx
christiaua: "More secure" 😂

client/src/components/LoginForm/SignupContent.jsx
```diff
@@ -2,61 +2,218 @@ import React, { Component } from 'react';
...
           />
           &nbsp;
           <EditText
             required
             label="Confirm password"
-            id="outlined-password-input"
+            id="confirm-password-input"
             type="password"
+            name="confirmPassword"
+            defaultValue={this.state.confirmPassword}
```
fionahang: You're missing the confirmPassword state in your constructor, so you get a warning when the user tries to type in the Confirm password textbox.

> Warning: A component is changing an uncontrolled input of type password to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components

PR #91, opened 2019-03-20
Yueh/frontend/refactor code

UAce:


Ty-Won:
@UAce this is the one commit on dev you don't have right?

Ty-Won:
Also there is another build failure on a backend test, not your fault, I'm trying to resolve it with backend team

PR #92, opened 2019-03-20
fixed stringify issue, querying for a student that does not exist issue

AlexLam1997:


backend/logic/users/users.js
```diff
@@ -150,6 +149,11 @@ var getStudentData = async studentID => {
     IN (SELECT offering_id, semester FROM student_course_offerings WHERE student_id = ?);`,
       [studentID]
     );
+
+    if(completedCourses.length == 0){
+      throw new Error("Student does not have any completed courses");
```
zerfuth: you should be able to not have any completed courses

PR #95, opened 2019-03-22
Tyrone/sign in/test

Ty-Won:
Added the tests to accompany the login endpoint.

### DIFF
--- a/backend/test/users_test.js
+++ b/backend/test/users_test.js
@@ -738,19 +738,13 @@ describe("Test get student completed courses", () => {
 });
 
 describe("Testing Login", () => {
-
-  let username ="team";
+  let username = "team";
   let password = "primus1234";
-  let email =  "testing@gmail.com";
-  let studentId = 250502459
+  let email = "testing@gmail.com";
+  let studentId = 250502459;
 
   before(async () => {
-    await users.insertStudentUser(
-      username,
-      password,
-      email,
-      studentId
-    );
+    await users.insertStudentUser(username, password, email, studentId);
   });
 
   it(`Null username responds with ${expectedUsernameNotEmpty}`, function(done) {
@@ -826,12 +820,17 @@ describe("Testing Login", () => {
   });
 
   it(`Long username greater than 64 length responds with ${expectedUsernameLessThan64}`, function(done) {
-    users.login("teammmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm", password).catch(error => {
-      return new Promise(function(resolve) {
-        assert.equal(error.message, expectedUsernameLessThan64);
-        resolve();
-      }).then(done);
-    });
+    users
+      .login(
+        "teammmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm",
+        password
+      )
+      .catch(error => {
+        return new Promise(function(resolve) {
+          assert.equal(error.message, expectedUsernameLessThan64);
+          resolve();
+        }).then(done);
+      });
   });
 
   it(`Short password less than 8 responds wth ${expectedPassword8To64}`, function(done) {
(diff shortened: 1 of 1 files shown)


backend/test/users_test.js
```diff
@@ -734,3 +737,101 @@ describe("Test get student completed courses", () => {
...
+    await users.insertStudentUser(
+      "team",
+      "primus1234",
+      "testing@gmail.com",
+      250502459
+    );
+  });
+
+  it(`Empty username responds with ${expectedUsernameNotEmpty}`, function(done) {
+    users.login(null, "primus1234").catch(error => {
```
zerfuth: Is null username the same as empty username? Like if you put in "" versus null, would it behave the same?

backend/test/users_test.js
```diff
@@ -734,3 +737,101 @@ describe("Test get student completed courses", () => {
...
+  it(`Empty password responds with ${expectedPasswordNotEmpty}`, function(done) {
+    users.login("team", null).catch(error => {
+      return new Promise(function(resolve) {
+        assert.equal(error.message, expectedPasswordNotEmpty);
+        resolve();
+      }).then(done);
+    });
+  });
+
+  it(`Empty fields responds with ${expectedPasswordNotEmpty}`, function(done) {
```
zerfuth: Did you mean responds with ${expectedUsernameNotEmpty} ?

backend/test/users_test.js
```diff
@@ -734,3 +737,101 @@ describe("Test get student completed courses", () => {
   });
 });
 
+describe("Testing Login", () => {
+  before(async () => {
+    await users.insertStudentUser(
+      "team",
+      "primus1234",
+      "testing@gmail.com",
+      250502459
```
si-x-li: Consider making these entries into variables so that we only have to modify at 1 location

backend/test/users_test.js
```diff
@@ -734,3 +737,101 @@ describe("Test get student completed courses", () => {
...
+    users.login(null, null).catch(error => {
+      return new Promise(function(resolve) {
+        assert.equal(error.message, expectedUsernameNotEmpty);
+        resolve();
+      }).then(done);
+    });
+  });
+
+  it(`Non existing Username responds with ${expectedNonExistentUser}`, function(done) {
+    users.login("teamm", "primus1234").catch(error => {
```
si-x-li: Same as above use variables instead

backend/test/users_test.js
```diff
@@ -734,3 +737,101 @@ describe("Test get student completed courses", () => {
...
+      return new Promise(function(resolve) {
+        assert.equal(error.message, expectedNonExistentUser);
+        resolve();
+      }).then(done);
+    });
+  });
+
+  it(`Incorrect Password responds with ${expectedIncorrectPassword}`, function(done) {
+    users.login("team", "primus123456789").catch(error => {
```
si-x-li: Same here

backend/test/users_test.js
si-x-li: I think it wouldn't behave the same since only null are caught in the validation

Ty-Won: Good catch, "" != null, the condition checks for 
userInfo == undefined || userInfo.length == 0
so it will catch it. I will add in a test for empty instead of null

backend/test/users_test.js
Ty-Won: thanks!

backend/test/users_test.js
```diff
@@ -734,3 +737,101 @@ describe("Test get student completed courses", () => {
...
+    users.login("team!$", "primus1234").catch(error => {
+      return new Promise(function(resolve) {
+        assert.equal(error.message, expectedInvalidUsername);
+        resolve();
+      }).then(done);
+    });
+  });
+
+  it(`Long username greater than 64 length responds with ${expectedUsernameLessThan64}`, function(done) {
+    users.login("teammmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm", "primus").catch(error => {
```
KevenLYF: lol

PR #96, opened 2019-03-22
Fixed validation error thing and inserted email in header and left dr…

zerfuth:
…awer

Still need to put the setEmail to login and redirect to homepage after finishing registration/login. StudentID could use similar approach or be set in Context


zerfuth:
There was merging conflict that i didnt properly handle. Please check again :)

PR #97, opened 2019-03-22
Added title and course desc to getStudentData

zerfuth:
There might be an issue with duplicate courses being retrieved but not 100% sure

### DIFF
--- a/backend/logic/users/users.js
+++ b/backend/logic/users/users.js
@@ -142,9 +142,11 @@ var getStudentData = async studentID => {
   let completedCourses, major, minors;
   try {
     completedCourses = await conn.query(
-      `SELECT course_code, semester
+    `SELECT course_offerings.course_code, course_offerings.semester, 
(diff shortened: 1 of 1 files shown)


si-x-li:
@zerfuth Are you using dev? There was some duplicate entries in course offerings

zerfuth:
That could be the reason why, yeah.

backend/logic/users/users.js
```diff
@@ -142,9 +142,11 @@ var getStudentData = async studentID => {
   let completedCourses, major, minors;
   try {
     completedCourses = await conn.query(
-      `SELECT course_code, semester
+    `SELECT course_offerings.course_code, course_offerings.semester,
```
KevenLYF: minor suggestion, consider using a variable to store a long query for better readability.

zerfuth: I attempted to indent it so it's still readable :')

PR #98, opened 2019-03-22
GET requests for courses and semesters

fionahang:
The user id is hardcoded to 999999999
We will also need to move the base url somewhere cleaner.

### DIFF
--- a/client/src/screens/CourseRegistration/CourseRegistration.js
+++ b/client/src/screens/CourseRegistration/CourseRegistration.js
@@ -56,6 +56,15 @@ class CourseRegistration extends Component {
   // Called on click of Select button, adds the course to list of selected courses
   onClickSelect() {
     this.setState(prevState => {
+      // Prevent adding the same course twice
+      if (
+        prevState.selectedCourses.some(
+          course => course.course_code === prevState.selectedSearch.course_code
+        )
+      ) {
+        return null;
+      }
+
       // Adds the selected semester to the course object
       const selection = {
         ...prevState.selectedSearch,
--- a/client/src/screens/CourseRegistration/CourseRegistration.js
+++ b/client/src/screens/CourseRegistration/CourseRegistration.js
@@ -56,7 +56,7 @@ class CourseRegistration extends Component {
   // Called on click of Select button, adds the course to list of selected courses
   onClickSelect() {
     this.setState(prevState => {
-      // Prevent adding the same course twice
+      // Prevents adding the same course twice
       if (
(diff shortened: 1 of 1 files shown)


client/src/screens/CourseRegistration/CourseRegistration.js
```diff
@@ -90,7 +72,62 @@ class CourseRegistration extends Component {
...
+      } else {
+        semesters.push('Winter ' + year.toString());
+        semesters.push('Fall ' + year.toString());
+      }
+    }
+    return semesters;
+  }
+
+  // TODO: remove because the student data should be passed down from login
+  fetchUserData() {
```
KevenLYF: it's an async call so it'd better to put async/await, otherwise u might get unexpected result.
```
fetchUserData = async() => {
    return await axios.get(
      'http://localhost:3001/users/getStudentData?studentID=999999999'
    );
  }
```

client/src/screens/CourseRegistration/CourseRegistration.js
```diff
@@ -90,7 +72,62 @@ class CourseRegistration extends Component {
...
+    return axios.get(
+      'http://localhost:3001/users/getStudentData?studentID=999999999'
+    );
+  }
+
+  fetchAllCourses() {
+    return axios.get('http://localhost:3001/courses/getAllCourses');
```
KevenLYF: really minor suggestion, consider using a global variable to store url.  This way, if url changed, you don't need to modify everywhere 
```
const url = 'http://localhost:3001/';
axios.get(`${url}courses/getAllCourses`);
```

client/src/screens/CourseRegistration/CourseRegistration.js
```diff
@@ -90,7 +72,62 @@ class CourseRegistration extends Component {
...
+            semesters: this.findSemesters(year),
+            allCourses: coursesData,
+          });
+        })
+      )
+      .catch(error => console.log('ERROR', error));
+  }
+
   render() {
+    const allCourses = this.state.allCourses;
```
KevenLYF: maybe you wanna do `object destructing` ? 
```
const { allCourses } = this.state;
```

client/src/screens/CourseRegistration/CourseRegistration.js
fionahang: the base url will be separated in a another file and exported so that it can be used everywhere. i'm going to leave my get request like this for now because it's easier to change once that is implemented

PR #101, opened 2019-03-22
Gareth/backend/create update major

GarethDPeters:


backend/logic/users/users.js
```diff
@@ -41,12 +57,10 @@ var deleteStudentUser = async username => {
...
   try {
     await connection.beginTransaction();
-    await connection.query(
-      `DELETE FROM students WHERE student_id =
-          (SELECT * FROM (SELECT student_id
-                          FROM students
-                          WHERE username = ?) t);`, 
-    [username]);
+    let student_id = await connection.query("SELECT student_id FROM students WHERE username = ?;", username);
+    await connection.query("DELETE FROM student_majors WHERE student_id = ?;", [student_id[0].student_id]);
```
si-x-li: The student's course offerings need to be deleted as well

backend/test/course_test.js
```diff
@@ -24,7 +21,10 @@ describe("Test retrieve course by tag", function() {
       "getCourseByTagTest",
       "getCourseByTagTest",
       "getCourseByTagTest@email.com",
-      260561055
+      260561055,
```
si-x-li: Put these as variables so that only 1 place needs to be modified in case something is changed

backend/validation/format.js
```diff
@@ -1,14 +1,15 @@
 let FormatError = require("./CustomErrors").FormatError;
 const MAX_USERNAME_LENGTH = 64;
 const MAX_PASSWORD_LENGTH = 64;
-const MAX_CURR_NAME_LENGTH = 128;
+const MAX_CURR_NAME_LENGTH = 256;
```
si-x-li: Check if the database matches the maximum length

backend/logic/users/users.js
```diff
@@ -277,12 +244,126 @@ var login = async (username, password) => {
   }
 };
 
+var assignStudentMajor = async (studentID, major) => {
+  let error = false;
```
KevenLYF:
```
let error;
```

backend/logic/users/users.js
```diff
@@ -277,12 +244,126 @@ var login = async (username, password) => {
...
+
+  if (!format.verifyStudentId(studentID)) {
+    error = "Invalid student id";
+  } else if (!format.verifyCurriculumName(major)) {
+    error = "Invalid curriculum name";
+  }
+
+  if (!error == false) {
```
KevenLYF:
```
if (error) {
...
}
```

backend/logic/users/users.js
```diff
@@ -277,12 +244,126 @@ var login = async (username, password) => {
...
+  let ifMajorExist;
+  let existingMajor;
+
+  let conn = await mysql.getNewConnection();
+  try {
+    ifUserExist = await conn.query(`SELECT COUNT(*) AS count FROM students WHERE student_id = ?;`, [studentID]);
+    ifMajorExist = await conn.query(`SELECT COUNT(*) AS count FROM curriculums WHERE curriculum_name = ?;`, [major]);
+    existingMajor = await conn.query(`SELECT COUNT(*) AS count FROM student_majors WHERE student_id = ?;`, [studentID]);
+  } catch (err) {
+    console.log(err);
```
KevenLYF: Minor suggestion, since you are creating a new connection later, you can put `conn.release()` into `catch` block, so that you don't have to do ```conn = await mysql.getNewConnection();``` again

backend/routes/users.js
```diff
@@ -168,4 +168,78 @@ router.get("/getStudentData", async (req, res) => {
...
+ * @author: Gareth Peters
+ *
+ */
+
+router.post("/assignStudentMajor", async (req, res) => {
+  const studentID = req.query.studentID;
+  const major = req.query.major;
+  try {
+    await users.assignStudentMajor(studentID, major);
+    res.status(200).send(true);
```
KevenLYF: @AlexLam1997  I think now we have to wrap the endpoint into `customResponse` right ?

PR #102, opened 2019-03-23
add cookies functionality for global access to studentid

dingding132:
Instead of using context, we can just use cookies. Only need to import the cookie library, see app.js and CurriculumDisplay for example. [Documentation](https://www.npmjs.com/package/universal-cookie)


client/src/App.js
```diff
@@ -9,13 +9,16 @@ import CourseRegistration from './screens/CourseRegistration';
...
   render() {
+    // TEMPORARY place to set student id, move this inside the login functionality
+    const cookies = new Cookies();
+    cookies.set('studentId', 123321123, { path: '/' });
+
     return (
-      <StudentProvider>
         <Router>
           <Switch>
             <Route path="/login" component={Login} />
```
UAce: The Route paths are outdated, you should have the paths to other pages

client/src/App.js
```diff
@@ -9,13 +9,16 @@ import CourseRegistration from './screens/CourseRegistration';
 import CourseSuggestion from './screens/CourseSuggestion';
 import Login from './screens/Login';
 import CurriculumDisplay from './screens/CurriculumDisplay';
-import { StudentProvider } from './contexts/StudentContext';
+import Cookies from 'universal-cookie';
```
UAce: I think we should just use Context

UAce: nvm apparently context is lost onRefresh :(

PR #108, opened 2019-03-23
Ua/frontend/use unified api

UAce:


client/src/components/RedirectError/RedirectError.js
```diff
@@ -0,0 +1,11 @@
+import history from '../../history';
+
+function RedirectError(error) {
+  console.error(error);
+  return history.push({
+    pathname: '/Error',
+    state: { error_code: error.response.status },
```
christiaua: `error.response.status` can be undefined

christiaua: `error.response.status` can be undefined

PR #116, opened 2019-03-24
Alex l/frontend/get curriculum years

AlexLam1997:


client/src/components/LoginForm/SignupContent.jsx
```diff
@@ -48,6 +49,19 @@ class SignupContent extends Component {
     );
   }
 
+  async getYears() {
+    try {
+      let url = 'curriculums/getCurriculumData';
+      let r = await Api().get(url);
```
Ty-Won: looks good, just wondering, I think front end team is using axios, try looking into that. Not 100% sure if the same package can be used to complete what you're trying to do.

package-lock.json
```diff
@@ -1,3 +1,48 @@
 {
-  "lockfileVersion": 1
+  "requires": true,
```
Ty-Won: Were there updates to a package.json?

AlexLam1997: I dont think so, i just ran an npm install after pulling from dev

PR #120, opened 2019-03-24
Get remaining courses and use real studentId

UAce:
…use sid from props

### DIFF
--- a/client/src/screens/CourseSuggestion/CourseSuggestion.js
+++ b/client/src/screens/CourseSuggestion/CourseSuggestion.js
@@ -32,18 +32,11 @@ class CourseSuggestion extends Component {
   }
 
   componentDidMount() {
-    this.setState(
-      {
-        studentId: this.props.studentId,
-      },
-      () => {
-        console.log('this is a state', this.state);
-        console.log('this is a prop', this.props.studentId);
-      }
-    );
+    this.setState({
+      studentId: this.props.studentId,
+    });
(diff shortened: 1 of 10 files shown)


client/src/screens/CourseSuggestion/CourseSuggestion.js
```diff
@@ -32,18 +32,11 @@ class CourseSuggestion extends Component {
...
-      {
-        studentId: this.props.studentId,
-      },
-      () => {
-        console.log('this is a state', this.state);
-        console.log('this is a prop', this.props.studentId);
-      }
-    );
+    this.setState({
+      studentId: this.props.studentId,
```
KevenLYF: `this.state.studentID` is not being referenced in this file, maybe we don't need this state. Just use `this.props.studentID`

client/src/screens/ErrorPage/ErrorPage.js
```diff
@@ -4,10 +4,13 @@ import WithHeaderBar from '../../hocs/WithHeaderBar';
 import './ErrorPage.css';
 
 class ErrorPage extends Component {
-  state = {};
   render() {
-    const code = this.props.location.state.error_code;
-    const message = this.props.location.state.error_message;
+    var code = '404';
```
KevenLYF:
```
    { location } = this.props;
    let code = '404';
    let message = 'Not Found';
    if (location.state) {
      code = location.state.error_code;
      message = location.state.error_message;
    }
```

